Issue #9, opened 2018-03-20
Squelch unused result warnings


jamwaffles:
I'm just being lazy now. The driver shold expose the result of transactions to the end user.

```
warning: unused `core::result::Result` which must be used
  --> src\interface\i2c.rs:23:9
   |
23 |         self.i2c.write(0x3c, &[0, cmd]);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(unused_must_use)] on by default

warning: unused `core::result::Result` which must be used
  --> src\interface\i2c.rs:39:9
   |
39 |         self.i2c.write(0x3c, &writebuf[0..buf.len() + 1]);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused `core::result::Result` which must be used
  --> src\interface\spi.rs:29:9
   |
29 |         self.spi.write(&[cmd]);
   |         ^^^^^^^^^^^^^^^^^^^^^^^

warning: unused `core::result::Result` which must be used
  --> src\interface\spi.rs:38:9
   |
38 |         self.spi.write(&buf);
   |         ^^^^^^^^^^^^^^^^^^^^^
```

PR #10, opened 2018-03-20
Error handling

scowcron:
Quick fix-up for #9 

* hal::blocking::spi::Transfer wasn't necessary on the SPI implementation, removed it.
* Added an Error associated type to the DisplayInterface trait.
* passed along errors where necessary rather than silencing them.


jamwaffles:
Looks good, thanks! Quick turnaround 👌.

Closes #9.

Issue #8, opened 2018-03-19
Documentation


jamwaffles:
Uncomment [this line](https://github.com/jamwaffles/ssd1306/blob/master/src/lib.rs#L3) and add some basic docs for everything.

jamwaffles:
Relevant discussion in #7

jamwaffles:
Some specific things I've noticed in the [rendered docs](https://jamwaffles.github.io/ssd1306/master/ssd1306/index.html):

* There should be a platform-agnostic example on the docs landing page with a graphics example, probably exercising all bits of the embedded_graphics API (triangle, square, circle, text, image). This will get updated with a following char mode example when #27 is ready. A link to a runnable example demonstrating the same for each display mode would be good too.
* To extend the previous point, there should be an explanation of the concept of display modes; this driver can't be used without choosing one!
* There should be better links between types/other language constructs in the docs
* Readme:
  * The image needs to go on a diet. It's huge at the moment.
  * `graphics` is on by default and no longer optional so should be de-listed
* The re-exports list needs an audit. This is related to #5 so can probably be fixed over there
* Should there be a description of how all the modules fit together?
* [`builder`](https://jamwaffles.github.io/ssd1306/master/ssd1306/builder/index.html) needs a couple of examples. I suggest:
  * A default/basic one that just sets up a display in graphics mode with SPI
  * One that exercises as many of the builder options as possible that changes the display size, rotation and I2C address in I2C mode.
* [`Rotate180`](https://jamwaffles.github.io/ssd1306/master/ssd1306/displayrotation/enum.DisplayRotation.html#variant.Rotate180) doesn't actually flip the display.
* [`interface`](https://jamwaffles.github.io/ssd1306/master/ssd1306/interface/index.html) should mention the [ridiculous](https://github.com/jamwaffles/ssd1306/blob/master/examples/blinky.rs#L45-L57) and [not so ridiculous](https://github.com/jamwaffles/ssd1306/blob/master/examples/blinky_i2c.rs#L40-L41) types required by RTFM apps.
* Each [mode](https://jamwaffles.github.io/ssd1306/master/ssd1306/mode/index.html#reexports) should have an explanation of what it does. Also worth mentioning that `RawMode` shouldn't be directly used and is only a helper for the builder.
* [`borrow_iface_mut`](https://jamwaffles.github.io/ssd1306/master/ssd1306/properties/struct.DisplayProperties.html#method.borrow_iface_mut) could use a short usage example.
[...]

PR #11, opened 2018-03-20
Add outline to docs and deny them in the future

scowcron:
Start to #8. This *DOES NOT* close that issue, the docs need to be more thoroughly fleshed out, but it allows us to enable the lint going forward and enforce good (developer) behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,13 @@
 // #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
 #![deny(warnings)]
+#![deny(missing_copy_implementations)]
+#![deny(trivial_casts)]
+#![deny(trivial_numeric_casts)]
+#![deny(unsafe_code)]
+#![deny(unstable_features)]
+#![deny(unused_import_braces)]
+#![deny(unused_qualifications)]
 
 #[cfg(feature = "graphics")]
 extern crate embedded_graphics;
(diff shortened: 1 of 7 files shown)


scowcron:
FWIW I'm not particularly keen on the Builder thing you've got going, those feel like they should just be functions rather than methods of some object.

scowcron:
scratch that, I see what you're doing, something like this. still feels a little clunky the way it is, will have to think about that.

``` rust
let builder = Builder::new().with_size(...);
```

src/lib.rs
```diff
@@ -1,14 +1,10 @@
...
 // TODO: Docs
-// #![deny(missing_docs)]
 // #![deny(missing_debug_implementations)]
-#![deny(missing_copy_implementations)]
-#![deny(trivial_casts)]
-#![deny(trivial_numeric_casts)]
-#![deny(unsafe_code)]
-#![deny(unstable_features)]
-#![deny(unused_import_braces)]
-#![deny(unused_qualifications)]
```
jamwaffles: Hmm. Can we keep these uncommented guards in? I don't think it's hurting anything at the moment. Rustfmt fixes the issues they bring up which is easy enough to run. Is there a reason you removed them?

scowcron: Yeah my bad. Was using a butcher's knife where I should have been using a scalpel.

jamwaffles:
Thanks 😁

PR #13, opened 2018-03-20
Use much less memory in I2C interface

### LINKED ISSUE #6: Fix duplicate buffer in I2C code
[`send_data()`](https://github.com/jamwaffles/ssd1306/blob/master/src/interface/i2c.rs#L28-L40) should be using less memory. @scowcron's [implementation](https://github.com/EdgewaterDevelopment/rust-ssd1306/blob/master/src/prelude.rs#L27) uses buffered chunks which might be a good way to go. Adding the initial `0x40` to the beginning of the buffer makes using another buffer a necessity at the moment.

jamwaffles:
Gone from 1025 byte buffer down to 17, although sending data to the display is marginally less efficient. Should be fine.

@scowcron This is from your `rust-ssd1306` crate. Any thoughts?

Fixes #6.

### DIFF
--- a/src/interface/i2c.rs
+++ b/src/interface/i2c.rs
@@ -27,20 +27,24 @@ where
         Ok(())
     }
 
-    // TODO: Send data in chunks to save memory. This code is particularly bad with 128x32 displays
-    // as half of `writebuf` is completely wasted.
     fn send_data(&mut self, buf: &[u8]) -> Result<(), I2C::Error> {
-        let mut writebuf: [u8; 1025] = [0; 1025];
+        let mut writebuf: [u8; 17] = [0; 17];
 
         // Data mode
         // 8.1.5.2 5) b) in the datasheet
         writebuf[0] = 0x40;
 
-        for (index, byte) in buf.iter().enumerate() {
-            writebuf[index + 1] = *byte;
+        // Noop if the data buffer is empty
+        if buf.is_empty() {
+            return Ok(());
         }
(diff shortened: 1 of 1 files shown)


scowcron:
other than the CI not working, looks good.

I was getting ready to submit a near identical PR. The old way would have failed on [stm32f30x](https://github.com/japaric/stm32f30x-hal/blob/d51a39b0ad1665c04ecead9ab148a5520dca139b/src/i2c.rs#L185) anyway.

I don't think there's anything to worry about sending an extra 64 bytes for every 1k at this scale. I'd rather eliminate the buffer entirely but the changes haven't come down yet that make that possible.

Ship it.

jamwaffles:
I'll remove CircleCI until we decide what to do with CI in general. Forgot I added it tbh.

Good catch on the stm32f30x. I'm using the stm32f103 so there's probably a bunch of stuff I'm missing.

I'm not too bothered about the memory usage. Maybe on an AVR or PIC but not on these ARM chips.

scowcron:
That `stm32f30x` is important because that's what japaric uses in all of his examples, which since he's the guy leading the charge means that's what a lot of people who try to pick this stuff up for the first time are going to be using. I have the f3 board he uses all the time, so I can be testing against that.

jamwaffles:
Yeah I've seen some of his neato accelerometer stuff he's done with that chip. It's good to have a range of test hardware!

PR #14, opened 2018-03-20
Un-hard-code I2C address

### LINKED ISSUE #12: Configurable I2C address
Either “0111100” (0x3C) or “0111101” (0x3D), can be selected as the slave address of SSD1306. We're hard-coded for 0x3C.

jamwaffles:
Section 8.1.5) a) in the datasheet states that the display address can be 0x3C or 0x3D, chosen with the D/C pin.

It can now be configured using the `.with_i2c_addr()` method on the `Builder`. Example:

```rust
let mut disp = Builder::new()
        .with_size(DisplaySize::Display128x32)
        .with_i2c_addr(0x3d)
        .connect_i2c(i2c);
```

Closes #12.

@scowcron This is where the builder pattern starts making sense to me; it feels more Rustic to chain a bunch of calls together to configure the object. Feels a bit like chaining iterator methods.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -11,6 +11,7 @@ use super::SSD1306;
 #[derive(Clone, Copy)]
 pub struct Builder {
     display_size: DisplaySize,
+    i2c_addr: u8,
 }
 
 impl Builder {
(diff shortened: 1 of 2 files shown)


scowcron:
Yep, I see where you're going with it. Seems a little bit odd that you can set an i2c address which will then go unused if you run the SPI route, but it's not really a big deal, and I suspect it will be optimized out with the --release flag anyway if someone decides to do that.

PR #17, opened 2018-03-22
Display rotation

jamwaffles:
Adds 0, 90, 180 and 270 degree rotations to the display. Usable like so:

```rust
let mut disp = Builder::new()
    .with_rotation(DisplayRotation::Rotate90)
    .connect_i2c(i2c);
```

Also adds the `get_dimensions()` method so consumers of this driver can position stuff relative to the display size.

### DIFF
--- a/examples/rotation_i2c.rs
+++ b/examples/rotation_i2c.rs
@@ -1,4 +1,4 @@
-//! Draw a 1 bit per pixel black and white image. On a 128x64 SSD1306 display over I2C.
+//! Draw the Rust logo centered on a 90 degree rotated 128x64px display
 //!
 //! Image was created with ImageMagick:
 //!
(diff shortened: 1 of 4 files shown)


examples/rotation_i2c.rs
```diff
@@ -0,0 +1,71 @@
...
+            duty_cycle: DutyCycle::Ratio1to1,
+        },
+        clocks,
+        &mut rcc.apb1,
+    );
+
+    let mut disp = Builder::new()
+        .with_rotation(DisplayRotation::Rotate90)
+        .connect_i2c(i2c);
+    disp.init();
```
therealprof: I smell new warnings here. ;)

jamwaffles:
Thanks for reviewing. Fixed the example warnings and added a `set_rotation()` method. Annoyingly it has to return a `Result` because it sets some config in the display. Makes the API a little messy but it's necessary.

jamwaffles:
Boss!

Issue #19, opened 2018-03-23
Setting a pixel out of bounds crashes the driver


jamwaffles:
The display buffer has a fixed length. There should be a guard that just does a noop if the pixel coordinates are out of bounds. I think this is better than returning a `Result::Err` because it allows the user more freedom in what they (try and) show on the display.

PR #21, opened 2018-03-23
Add guard around out of bounds pixel set

jamwaffles:
Closeses #19 

@scowcron I can't decide if this should be a noop or return a `Result`. The latter feels more Rustic, but what's the user going to do with it? They should be checking the display dimensions, not whether a pixel was out of bounds or not. Thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,17 +82,22 @@ where
         }
     }
 
-    /// Turn a pixel on or off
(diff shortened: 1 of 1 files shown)


jamwaffles:
@scowcron merging to get this moving. It can be changed to return an `Error` in the future if people decide that's a better idea.

PR #24, opened 2018-03-28
Silence warnings in examples

therealprof:

### DIFF
--- a/examples/image_i2c.rs
+++ b/examples/image_i2c.rs
@@ -50,12 +50,12 @@ fn main() {
     );
 
     let mut disp = Builder::new().connect_i2c(i2c);
-    disp.init();
-    disp.flush();
+    disp.init().unwrap();
+    disp.flush().unwrap();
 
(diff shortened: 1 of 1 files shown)


therealprof:
I hate panics and I hate warnings so I didn't want to introduce the former just to cure the latter. ;)

jamwaffles:
Ha, I hate both too, but I think in the examples' case I'd rather it panicked. I think it's easier to pick up what went wrong with a debugger than just silently(?) continuing with an unused var, and the program can't really do much else at that point. Would you mind changing them to `.unwrap()`s?

therealprof:
I hate debuggers, too. 😉 But I'll `unwrap()` it for you.

therealprof:
Done.

jamwaffles:
There's no pleasing some people 😂 but thanks for the changes 😃. The other examples should get updated but let's do that as we come to them.

therealprof:
The other examples were fine already.

PR #29, opened 2018-03-29
Travis

jamwaffles:
Build core and examples in Travis. Auto docs upload will come at a later date.


jamwaffles:
@therealprof Only the basics right now. I'd like to add format checking (I tried just now but my Windows line endings seem to be screwing things up - will try from Mac later) and doc publish at some point, but this a good start :)

Issue #30, opened 2018-03-30
"no method named `draw` found for type"


chocol4te:
I am trying to use this library in my project with little success. When I compile my code as an example it works without issue, but in a standalone project I get the following error:

`error[E0599]: no method named ``draw`` found for type ``ssd1306::SSD1306<ssd1306::interface::I2cInterface<hal::i2c::I2c<hal::<unnamed>::I2C1, (hal::gpio::gpiob::PB6<hal::gpio::Alternate<hal::gpio::OpenDrain>>, hal::gpio::gpiob::PB7<hal::gpio::Alternate<hal::gpio::OpenDrain>>)>>>`` in the current scope
  --> src/main.rs:62:10
   |
62 |     disp.draw(Font6x8::render_str("Hello world!", (0, 0)).into_iter());
   |          ^^^^
`

`error[E0599]: no method named ``draw`` found for type ``ssd1306::SSD1306<ssd1306::interface::I2cInterface<hal::i2c::I2c<hal::<unnamed>::I2C1, (hal::gpio::gpiob::PB6<hal::gpio::Alternate<hal::gpio::OpenDrain>>, hal::gpio::gpiob::PB7<hal::gpio::Alternate<hal::gpio::OpenDrain>>)>>>`` in the current scope
  --> src/main.rs:63:10
   |
63 |     disp.draw(Font6x8::render_str("Hello Rust!", (0, 16)).into_iter());
   |          ^^^^
`

Any idea what it could be?

Thanks!

jamwaffles:
Documentation is pretty thin at the moment, sorry about that! Try adding
`--features graphics` to your Cargo command. It will include the
embedded_graphics crate which adds that method.

therealprof:
Have I mentioned that having the main graphics driver depend on a feature-gate is not very user-friendly? 😉

jamwaffles:
You have indeed 😁 what do you say about enabling the `graphics` feature by default, but leave it in as a feature so it can be disabled? Not sure how useful that is or if I'm just bikeshedding here.

therealprof:
Well, as I said, IMHO it should either be always on or we supplement the project with a simpler alternative (i.e. no external dependency, like the option we have on hand from the third contributer 😉) and use that as default. I think it makes sense to have graphics enabled by default but keep it as a feature if someone absolutely wants to get rid of the dependency.

chocol4te:
[...]

PR #31, opened 2018-03-31
Turn graphics feature on by default

jamwaffles:
As raised by @therealprof and @chocol4te in #30, it makes sense to have
these features enabled by default, but disable-able if someone _really_
needs to. The compiler will optimise it all out anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,5 +44,5 @@ debug = true
 lto = true
 
 [features]
-default = []
+default = ["graphics"]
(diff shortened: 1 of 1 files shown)


jamwaffles:
Cheers. By the way, are you allowed to merge into master? Not sure if I've set the Github permissions up right.

therealprof:
Seems so. ;)

chocol4te:
Works great! :)

PR #37, opened 2018-04-01
Refactored some code into DisplayProperties to simplify mode handling

therealprof:

### DIFF
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -1,20 +1,22 @@
 //! Buffered display module for use with the embedded_graphics crate
 
 use displaysize::DisplaySize;
+use displayrotation::DisplayRotation;
 
-pub use displayrotation::DisplayRotation;
 use command::{AddrMode, Command, VcomhLevel};
 
 use hal::blocking::delay::DelayMs;
 use hal::digital::OutputPin;
 use interface::DisplayInterface;
+use properties::DisplayProperties;
 
 /// GraphicsMode
-pub struct GraphicsMode<DI> {
-    iface: DI,
+pub struct GraphicsMode<DI>
+where
+    DI: DisplayInterface,
+{
+    properties: DisplayProperties<DI>,
     buffer: [u8; 1024],
-    display_size: DisplaySize,
-    display_rotation: DisplayRotation,
 }
 
 impl<DI> GraphicsMode<DI>
@@ -22,15 +24,9 @@ where
     DI: DisplayInterface,
 {
     /// Create new GraphicsMode instance
-    pub fn new(
-        iface: DI,
-        display_size: DisplaySize,
-        display_rotation: DisplayRotation,
-    ) -> GraphicsMode<DI> {
+    pub fn new(properties: DisplayProperties<DI>) -> Self {
         GraphicsMode {
-            iface,
-            display_size,
-            display_rotation,
+            properties,
             buffer: [0; 1024],
         }
     }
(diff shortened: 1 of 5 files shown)


therealprof:
@jamwaffles In order to move between display modes more easily (and to factor out the common code) we need some sort of intermediate structure in between. If you have a better idea for a name or implementation, I'd like to hear it.

src/mode/graphics.rs
```diff
@@ -113,75 +113,54 @@ where
...
-            DisplayRotation::Rotate0 | DisplayRotation::Rotate180 => (w, h),
-            DisplayRotation::Rotate90 | DisplayRotation::Rotate270 => (h, w),
+        match display_rotation {
+            &DisplayRotation::Rotate0 | &DisplayRotation::Rotate180 => (w, h),
+            &DisplayRotation::Rotate90 | &DisplayRotation::Rotate270 => (h, w),
         }
     }
 
     /// Set the display rotation
     pub fn set_rotation(&mut self, rot: DisplayRotation) -> Result<(), DI::Error> {
```
jamwaffles: Can we put common methods like this into a `DisplayMode` (or something) trait that every mode implements? It would make consistent, common behaviour easier to handle I think.

src/mode/graphics.rs
```diff
@@ -55,29 +51,33 @@ where
 
     /// Write out data to display
     pub fn flush(&mut self) -> Result<(), DI::Error> {
-        let (display_width, display_height) = self.display_size.dimensions();
+        let display_size = self.properties.get_size().clone();
```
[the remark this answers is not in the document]
therealprof: I can try but I think this'll cause borrowed while mut borrowed kind of errors.

src/mode/graphics.rs
```diff
@@ -113,75 +113,54 @@ where
...
+        Command::DisplayOffset(0).send(iface)?;
+        Command::StartLine(0).send(iface)?;
         // TODO: Ability to turn charge pump on/off
-        Command::ChargePump(true).send(&mut self.iface)?;
-        Command::AddressMode(AddrMode::Horizontal).send(&mut self.iface)?;
+        Command::ChargePump(true).send(iface)?;
+        Command::AddressMode(AddrMode::Horizontal).send(iface)?;
 
-        self.configure_rotation()?;
+        //self.configure_rotation()?;
```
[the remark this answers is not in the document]
therealprof: Thanks, I missed that. However that madness is supposed to go away with a proper initialisation interface.

src/mode/graphics.rs
therealprof: Yes, absolutely. This new layer is supposed to enable exactly that. I tried a few ways with traits to just get the common stuff done but some problems I ran into were:
- Switching mode requires releasing all resources and reinitialising it in the new mode (which is not only superfluous and also has a lot of repetition in it and requires a lot of changes if we add something)
- It's not easily possibility to add custom data (e.g. the buffer) to modes requiring it

By putting all common aspects in a custom layer it should be much more easily possible to create the desired DisplayMode trait that provides a common interface allowing to switch between drivers on the fly without having to care about a bucket load of crazy trait dependencies and bounds. Still need to get that 20+ years of C/C++ experience out of my brain... ;)

therealprof:
To address the cloning I had to go exactly the other around and turn the returned references into non-references so they would be implicitly be copied and thus not clash with `mut` borrows. Those are only temporary anyway since there should be a proper setup interface and all the magic happen inside the new layer instead of in the display mode.

jamwaffles:
Yeah, the changed return behaviour looks lots better now! Thanks for the change 👍

src/mode/graphics.rs
jamwaffles: Ok, if you've got plans to clean this up I'm looking forward to seeing them :). Releasing and reinitialising sounds painful so let's not do that haha.

jamwaffles:
Adds a bit of complexity, but as far as I can see there's no change to the public interface so I'll merge and we can clean the code up more in the future.

src/mode/graphics.rs
therealprof: Yeah, this stuff really shouldn't be there. Planning to move everything out except for mode specific initialisation, cleanup and stuff.

PR #39, opened 2018-04-02
Moved more general code to properties.rs for deduplication

therealprof:

### DIFF
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -111,8 +111,8 @@ where
         }
     }
 
-    // Display is set up in column mode, i.e. a byte walks down a column of 8 pixels from column 0 on the left, to column _n_ on the right
-    /// Initialize display in column mode.
+    /// Display is set up in column mode, i.e. a byte walks down a column of 8 pixels from
+    /// column 0 on the left, to column _n_ on the right
     pub fn init(&mut self) -> Result<(), DI::Error> {
         self.properties.init_column_mode()?;
         Ok(())
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -1,12 +1,12 @@
 //! Buffered display module for use with the embedded_graphics crate
 
-use displaysize::DisplaySize;
-use displayrotation::DisplayRotation;
-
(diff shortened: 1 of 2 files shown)


therealprof:
Moving some common code from graphics mode over to properties.rs for deduplication purposes and to clean up the mode. The whole chunk of `init_column_mode()` should really be broken up into more sizeable and reusable pieces so we don't need to have another chunk for a slight variation of this. Also the rest of the `Command` should be moved out into nice functions and `send_data()` on the borrowed interface be properly encapsulated in properties.rs.

src/properties.rs
```diff
@@ -30,6 +31,43 @@ where
         }
     }
 
+    /// Initialize display in column mode
+    pub fn init_column_mode(&mut self) -> Result<(), DI::Error> {
+        // TODO: Break up into nice bits so display modes can pick whathever they need
+        // Display is set up in column mode, i.e. a byte walks down a column of 8 pixels from
```
jamwaffles: Can you move this to the function docblock? Would be helpful to have in the rendered docs.

src/mode/graphics.rs
```diff
@@ -113,52 +114,13 @@ where
     // Display is set up in column mode, i.e. a byte walks down a column of 8 pixels from column 0 on the left, to column _n_ on the right
```
jamwaffles: This would be useful to have in the rendered docs too.

src/mode/graphics.rs
```diff
@@ -37,6 +37,7 @@ where
     }
 
     /// Reset display
+    // TODO: Move to a more appropriate place
```
jamwaffles: Hmm, yes this is a strange one. It's only applicable to the SPI interface which makes placing it in the architecture of this crate pretty weird. Do you have any thoughts where this could go?

src/properties.rs
therealprof: I agree, that's why I was asking for naming ideas before. `DisplaySetup` would work, `DisplayConfiguration` also.

jamwaffles:
Thanks for the doc changes. I can't think of a good name now so let's just get this change merged :)

PR #46, opened 2018-04-08
Move back to Cargo

jamwaffles:
Based on [this PSA](https://users.rust-lang.org/t/psa-breaking-change-extern-crate-compiler-builtins-is-now-included-in-no-std-crates/16704) and [this PSA](https://users.rust-lang.org/t/psa-you-no-longer-need-xargo-to-do-arm-cortex-m-development/16703) hopefully we no longer need to use Xargo for builds in Travis.

This also requires that users of this crate use `nightly-2018-04-08` or later to be able to use Cargo instead of Xargo, although both should continue to work.


.travis.yml
```diff
@@ -4,7 +4,7 @@
 dist: trusty
 language: rust
 rust:
-  - nightly
+  - nightly-2018-04-08
```
therealprof: Hm, this seems dubious but otherwise okay.

jamwaffles: Yeah let me fix that, just spotted too

jamwaffles:
Fixed! Can you approve again please? Thanks 😄

PR #49, opened 2018-04-08
Features/refactorcommands

therealprof:
Probably not the best name in the world. This refactors the direct use of the interface in the graphicsmode to only use methods provided by the `DisplayProperties` and removes the `borrow_iface_mut()`.

### DIFF
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -3,8 +3,6 @@
 use hal::blocking::delay::DelayMs;
 use hal::digital::OutputPin;
 
-use command::Command;
-
 use displayrotation::DisplayRotation;
 use displaysize::DisplaySize;
 use interface::DisplayInterface;
@@ -65,17 +63,16 @@ where
     /// Write out data to display
     pub fn flush(&mut self) -> Result<(), ()> {
         let display_size = self.properties.get_size();
-        let iface = self.properties.borrow_iface_mut();
 
+        // Ensure the display buffer is at the origin of the display before we send the full frame
+        // to prevent accidental offsets
         let (display_width, display_height) = display_size.dimensions();
-
-        Command::ColumnAddress(0, display_width - 1).send(iface)?;
-        Command::PageAddress(0.into(), (display_height - 1).into()).send(iface)?;
+        self.properties.set_render_area((0, display_width), (0, display_height))?;
 
         match display_size {
-            DisplaySize::Display128x64 => iface.send_data(&self.buffer),
-            DisplaySize::Display128x32 => iface.send_data(&self.buffer[0..512]),
-            DisplaySize::Display96x16 => iface.send_data(&self.buffer[0..192]),
+            DisplaySize::Display128x64 => self.properties.render(&self.buffer),
+            DisplaySize::Display128x32 => self.properties.render(&self.buffer[0..512]),
+            DisplaySize::Display96x16 => self.properties.render(&self.buffer[0..192]),
         }
     }
 
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -67,12 +67,12 @@ where
         // Ensure the display buffer is at the origin of the display before we send the full frame
         // to prevent accidental offsets
         let (display_width, display_height) = display_size.dimensions();
-        self.properties.set_render_area((0, display_width), (0, display_height))?;
+        self.properties.set_draw_area((0, display_width), (0, display_height))?;
 
         match display_size {
-            DisplaySize::Display128x64 => self.properties.render(&self.buffer),
-            DisplaySize::Display128x32 => self.properties.render(&self.buffer[0..512]),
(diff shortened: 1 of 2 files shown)


src/mode/graphics.rs
```diff
@@ -65,17 +63,15 @@ where
     /// Write out data to display
     pub fn flush(&mut self) -> Result<(), ()> {
         let display_size = self.properties.get_size();
-        let iface = self.properties.borrow_iface_mut();
 
+        // TODO: Why is this here? Once during initialisation should suffice?
```
jamwaffles: Correct, I'm pretty sure the chip will remember these settings until told otherwise. Move them to init and see what happens? Presumably charmode will want to set this dynamically at some point to support text positioning?

src/properties.rs
```diff
@@ -67,9 +67,17 @@ where
...
-        &mut self.iface
+    /// Set the area on the display the data should be rendered onto
+    pub fn set_render_area(&mut self, start: (u8, u8), end: (u8, u8)) -> Result<(), ()> {
+        Command::ColumnAddress(start.0, start.1 - 1).send(&mut self.iface)?;
+        Command::PageAddress(start.1.into(), (end.1 - 1).into()).send(&mut self.iface)?;
+        Ok(())
+    }
+
+    /// Render data onto the display
+    pub fn render(&mut self, buffer: &[u8]) -> Result<(), ()> {
```
jamwaffles: Can we call this `draw()` or something? We're not really rendering anything, just sending a bunch of data to the display. It would also be worth mentioning in the doc comment that what is drawn depends on how `set_render_area()` is set.

src/mode/graphics.rs
```diff
@@ -65,17 +63,15 @@ where
...
-        Command::PageAddress(0.into(), (display_height - 1).into()).send(iface)?;
+        self.properties.set_render_area((0, display_width), (0, display_height))?;
 
         match display_size {
-            DisplaySize::Display128x64 => iface.send_data(&self.buffer),
-            DisplaySize::Display128x32 => iface.send_data(&self.buffer[0..512]),
-            DisplaySize::Display96x16 => iface.send_data(&self.buffer[0..192]),
+            DisplaySize::Display128x64 => self.properties.render(&self.buffer),
+            DisplaySize::Display128x32 => self.properties.render(&self.buffer[0..512]),
+            DisplaySize::Display96x16 => self.properties.render(&self.buffer[0..192]),
```
jamwaffles: This looks much cleaner

src/properties.rs
therealprof: Certainly.

src/mode/graphics.rs
therealprof: Surely charmode will use the very same on the fly when cursor repositioning is requested. It's a readily available method so I don't see a problem with that. Only the static call on every flush is a bit weird, if we always draw a full frame the display should always end up at the origin. Maybe we should throw in some error handling to reset the position in case a frame could not be fully transmitted but I'm not sure what will happen then anyways...

therealprof:
Better now?

src/mode/graphics.rs
jamwaffles: That's what I was getting at; I think it should live in `init()` for graphics mode, but char mode will probably make copious use of it elsewhere :). If some code changes the position somewhere, then the user has the responsibility to change it back for full screen rendering to work properly IMO. I'm envisaging scenarios where higher framerates are more important than correctness for graphics mode, so the less we do in `flush()` the better.

therealprof: ? This change factors that into the `set_draw_area()` method which can easily be used from `init()` or on the fly by charmode if reposititioning of the cursor is requested. What do you see missing here?

jamwaffles: I think we're misunderstanding each other :). I'm agreeing with your code comment; calling `set_draw_area()` here in `flush()` is just a bit of safety and can probably be moved to `init()`. That said it might be worth leaving in `flush()` as, like you said, it ensures the framebuffer is drawn in the right position. It's safety vs speed. I'd rather have `set_draw_area()` called once in `init()` for speed reasons, but what do you think?

therealprof: My 2 points (worth a cent each):
- There're two possibilities why the local frame buffer position might end up not in line with the display buffer: A (temporary) communication glitch or someone called `set_draw_area()` manually
- Per design the current graphics mode always does full redraws and the reset of the draw area really only adds a few instructions and bytes to be sent over the interface so the performance implications should be negligible compared to sending the full buffer so I'd probably leave it in for simplicity reasons.

jamwaffles: Ok, thanks for explaining :). Your points make sense. Can you remove the TODO comment and rebase, then I'll get this merged?

therealprof: Done, I've replaced it with a comment explaining why we do it with every `flush()`.

jamwaffles:
Danke schön!

PR #51, opened 2018-04-09
Doc improvements

### LINKED ISSUE #8: Documentation

jamwaffles:
From https://github.com/jamwaffles/ssd1306/issues/8#issuecomment-379584031:

* [x] There should be a platform-agnostic example on the docs landing page with a graphics example, probably exercising all bits of the embedded_graphics API (triangle, square, circle, text, image). This will get updated with a following char mode example when #27 is ready. A link to a runnable example demonstrating the same for each display mode would be good too.
* [x] To extend the previous point, there should be an explanation of the concept of display modes; this driver can't be used without choosing one!
* [x] There should be better links between types/other language constructs in the docs
* [x] Readme:
  * [x] 5d0d131 The image needs to go on a diet. It's huge at the moment.
  * [x] `graphics` is on by default and no longer optional so should be de-listed
* ~The re-exports list needs an audit. This is related to #5 so can probably be fixed over there~ will be fixed in #5
  * ~Should there be a description of how all the modules fit together?~
* [x] [`builder`](https://jamwaffles.github.io/ssd1306/master/ssd1306/builder/index.html) needs a couple of examples. I suggest:
  * [x] A default/basic one that just sets up a display in graphics mode with SPI
  * [x] One that exercises as many of the builder options as possible that changes the display size, rotation and I2C address in I2C mode.
* [x] 7aef625a4916a79f934cf20b27d269d1d9a5c471 [`Rotate180`](https://jamwaffles.github.io/ssd1306/master/ssd1306/displayrotation/enum.DisplayRotation.html#variant.Rotate180) doesn't actually flip the display.
* [x] [`interface`](https://jamwaffles.github.io/ssd1306/master/ssd1306/interface/index.html) should mention the [ridiculous](https://github.com/jamwaffles/ssd1306/blob/master/examples/blinky.rs#L45-L57) and [not so ridiculous](https://github.com/jamwaffles/ssd1306/blob/master/examples/blinky_i2c.rs#L40-L41) types required by RTFM apps.
* [x] Each [mode](https://jamwaffles.github.io/ssd1306/master/ssd1306/mode/index.html#reexports) should have an explanation of what it does. Also worth mentioning that `RawMode` shouldn't be directly used and is only a helper for the builder.
* [x] Make sure each module has a more-than-one-line description of what it does, preferably with a code example.
* [x]  TerminalMode docs now that #27 is merged
  * [x] Main page example
  * [x] More comprehensive docs for its page
* [x] Add target device to each example
* [x] Mode examples
  * [x] Graphics mode
  * [x] Terminal mode

Closes #8


jamwaffles:
@therealprof I think this is ready to go. Might be some stuff I haven't covered/could expand on so let me know if I should add anything else, particularly around `TerminalMode`.

jamwaffles:
Thanks! Yep, always round 2, 3, 4, ... if we need it :D

PR #52, opened 2018-04-09
Fixed bug recently introduced in set_draw_area()

therealprof:


therealprof:
I'm sorry but I introduced a bug which prevents charmode from working with my latest refactoring and potentially negative side effects for graphicsmode, too.

jamwaffles:
Woops! I should've caught this in review. Thanks for fixing :)

PR #54, opened 2018-04-17
Use LLD to link programs

jamwaffles:
[We can now use LLD to link Cortex-M programs!](https://users.rust-lang.org/t/cortex-m-rt-v0-4-0-now-you-can-link-arm-cortex-m-programs-using-lld/16751). This change moves over to using lld in `.cargo/config`. I haven't tested this with any hardware yet so don't merge just yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,7 @@ cortex-m = "0.4.3"
 aligned = "0.1.1"
 cortex-m-rtfm = "0.3.2"
 cortex-m-rtfm-macros = "0.3.1"
-cortex-m-semihosting = "0.2.0"
 cortex-m-rt = "0.4.0"
-panic-semihosting = "0.1.0"
 panic-abort = "0.1.0"
 
 # Must be installed from Git currently, not sure why it's not on crates.io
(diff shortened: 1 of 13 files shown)


jamwaffles:
@therealprof this is working fine for me, can you give it a test and merge in if it all looks good?

Cargo.toml
```diff
@@ -17,12 +17,12 @@ optional = true
...
 aligned = "0.1.1"
-cortex-m-rtfm = "0.3.1"
-cortex-m-rtfm-macros = "0.3.0"
-
-[dev-dependencies.cortex-m-rt]
-features = ["abort-on-panic"]
-version = "0.3.12"
+cortex-m-rtfm = "0.3.2"
+cortex-m-rtfm-macros = "0.3.1"
+cortex-m-semihosting = "0.2.0"
```
therealprof: Why the new semihosting dependency?

jamwaffles: Just some testing cruft I think, removed in db92a41.

jamwaffles:
Updated from master, merge at your leisure :)

PR #55, opened 2018-04-18
Fix typo in set_draw_area()

jamwaffles:
@therealprof None of my stuff was working. I thought I was going insane, turns out it was just a typo in argument order 😂. TerminalMode [does it right](https://github.com/jamwaffles/ssd1306/blob/fix-graphics-addressing/src/mode/terminal.rs#L162).


therealprof:
Woops, sorry, I though I fixed that.

PR #61, opened 2018-04-29
Add support for 90 and 270 degree text rotation

### LINKED ISSUE #28: Rotation is all screwy
From #27:

> Rotation acts really funny for 90 and 270 degrees, it basically mirrors the characters and changes from l-t-r to r-t-l mode with character mode.

This is probably an artifact of me only testing an image (of the Rust logo). I'll do some more thorough testing taking into account LTR/RTL direction. I think it's fine for graphics where entire display is refreshed at once, but char mode will depend on fill/scroll directions.

jamwaffles:
Closes #28 

This PR adds support for non-screwy 90 and 270 degree rotations in terminal mode. There's a performance hit when using these rotation values, however, as the byte array returned by `to_bitmap()` has to be transmuted. I was wondering about just storing another array of chars with the transmute preapplied, but we'd be using double the memory in dev mode. It'd probably get optimised out in release mode which is worth something. 

I also have to keep track of where the char is and use `set_draw_area()` appropriately otherwise the text ends up looking like this:

img_20180429_185243

I couldn't see anything in the datasheet that would help out here unfortunately.

### DIFF
--- a/src/mode/terminal.rs
+++ b/src/mode/terminal.rs
@@ -143,6 +143,7 @@ where
 /// Terminal mode handler
 pub struct TerminalMode<DI> {
     properties: DisplayProperties<DI>,
+    pos: u32,
 }
 
 impl<DI> DisplayModeTrait<DI> for TerminalMode<DI>
@@ -151,7 +152,7 @@ where
 {
     /// Create new TerminalMode instance
     fn new(properties: DisplayProperties<DI>) -> Self {
-        TerminalMode { properties }
+        TerminalMode { properties, pos: 0 }
     }
 
     /// Release all resources used by TerminalMode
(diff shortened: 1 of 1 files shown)


therealprof:
Hm, that is nice. But I'm wondering whether we should leave it (it being rotation) for "Enhanced TerminalMode" which will have to keep the position anyway to properly react to special character and possibly implement the line shifting (instead of wrapping) and keep this one simple.

jamwaffles:
Yeah, I'm not happy that this adds quite a bit of complexity to the mode. Having it be stateless (i.e. no `set_draw_area()`) is really nice. I'll make a new PR which just adds some doc text explaining that terminal mode doesn't support 90 and 270º rotations to replace this one.

PR #68, opened 2018-05-15
proper fix for stm32f103xx-hal blocking I2C changes

ilya-epifanov:


jamwaffles:
I'm mostly ok with it because it's for the example code, although it definitely looks weird and increases the friction around actually using this crate.

PR #74, opened 2019-01-03
Don't wrap around when x >= width

WGH-:
`set_pixel` already does nothing if pixel memory index is larger than buffer size.

However, when pixel overflows by x, it simply overflows into the next row (y coordinate) instead of being ignored.

This commit fixes it, and causes it to be ignored.


jamwaffles:
This is published in version [0.2.2](https://crates.io/crates/ssd1306/0.2.2). Give it a try and let me know how it works out for you.

WGH-:
I tested the change before making a PR, if that's what you are asking :)

Issue #76, opened 2019-03-09
Use with a ssd1305 chip, and display size 132x64.


everett1992:
I have [this display](https://www.adafruit.com/product/2675) from adafruit. It's a ssd1305 but in adafruit's circuit python tutorials they use a ssd1306 driver. I ran out of memory programming the chip in python so I'm switching to rust and found this library.

In python I had to define the display size as [132x64](https://github.com/adafruit/Adafruit_CircuitPython_SSD1306/issues/24), despite being advertised as 128x32. With this display size the top left pixel is 32x4. If I use a display size of 128x32 then every other line is disabled (looks like an old crt).

I see the same behavior with your library, if I use a TerminalMode driver with DisplaySize::Display128x32 then every other line is disabled. If I use 128x64 the display is much better, each character looks correct, but the text starts off screen. The last columns of the display also do not work, it will show random pixels when the display is initialized and they will not change or clear.

 - Would it make sense to add `Display132x64`?
 - I saw `display.properties.set_display_area` but it isn't public, is there a way to set that?

everett1992:
I closed the PR because I think it's too heavy weight right now.

I reverted all changes in the PR back to master. 
Here is my display code, and photo of the display using the crate. 
```rust
 let mut disp: GraphicsMode<_> = Builder::new()                              
         .with_size(DisplaySize::Display128x32)                                                                     
         .connect_spi(spi, dc)                                              
         .into();                                                           
                                                                            
    disp.reset(&mut rst, &mut delay);                                       
    disp.init().unwrap();                                                   
    disp.clear();                                                           
                                                                            
    disp.draw(                                                              
        Rect::new(Coord::new(0,0), Coord::new(127,31))                      
        .with_stroke(Some(1u8.into()))                                      
        .with_fill(Some(0u8.into()))                                        
        .into_iter()                                                        
    );                                                                                                     
                                                                                
    disp.flush().unwrap();                                                      
    loop {}              
```

[...]

Issue #20, opened 2018-03-23
More efficient display buffer


jamwaffles:
After #19, there should be a way of giving the driver a different length display buffer depending on the size of the screen.

jamwaffles:
After some initial investigation, this probably needs const generics to be in stable before being releasable. Something could be achieved with the `typenum` crate, but that's a kludge imo. Const generics are in nightly so it shouldn't be long until it lands.

jamwaffles:
Blocked on rust-lang/rust#44580

PR #77, opened 2019-03-09
Add 132x64 display size for some ssd1305 displays

everett1992:
starts #76

### DIFF
--- a/src/displaysize.rs
+++ b/src/displaysize.rs
@@ -25,4 +25,26 @@ impl DisplaySize {
             DisplaySize::Display96x16 => (96, 16),
         }
     }
+
+    /// Get the number of characters this display can show. based on a 8x8 font.
+    pub fn numchars(&self) -> u8 {
+        match *self {
+            DisplaySize::Display128x64 => 128,
+            DisplaySize::Display132x64 => 64,
+            DisplaySize::Display128x32 => 64,
+            DisplaySize::Display96x16 => 24,
+        }
+    }
+
+    /// Get the area within the display that characters will be drawn.
+    /// This is primaritly implemented for the framebuffer 132x64 hack
+    /// because the buffer is larger than the draw area.
+    pub fn draw_area(&self) -> ((u8, u8), (u8, u8)) {
+        match *self {
+            DisplaySize::Display128x64 => ((0, 0), (128, 64)),
+            DisplaySize::Display132x64 => ((4, 32), (132, 64)),
+            DisplaySize::Display128x32 => ((0, 0), (128, 32)),
+            DisplaySize::Display96x16 => ((0, 0), (96, 16)),
+        }
+    }
 }
--- a/src/displaysize.rs
+++ b/src/displaysize.rs
@@ -6,6 +6,8 @@
 pub enum DisplaySize {
     /// 128 by 64 pixels
     Display128x64,
+    /// 132 by 64 pixels, try this for ssd1305 chips that are 128x64.
+    Display132x64,
     /// 128 by 32 pixels
     Display128x32,
     /// 96 by 16 pixels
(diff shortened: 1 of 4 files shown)


src/mode/graphics.rs
```diff
@@ -85,8 +85,11 @@ where
         self.properties
             .set_draw_area((0, 0), (display_width, display_height))?;
 
+        // XXX Might need to change the draw area on 132x64?
+
         match display_size {
             DisplaySize::Display128x64 => self.properties.draw(&self.buffer),
+            DisplaySize::Display132x64 => self.properties.draw(&self.buffer[0..1056]),
```
everett1992: These numbers look like width*height/8, but I'm not sure if I'm right, or if some displays would need a different value. For 132x64 I don't know if we need to draw at an offset like [offset..offset+1056]

This wholeblock could be switched to this
```suggestion
self.properties.draw(&self.buffer[0..(display_width*display_height / 8)]);
```

src/mode/terminal.rs
```diff
@@ -170,14 +170,15 @@ where
 
         let numchars = match display_size {
             DisplaySize::Display128x64 => 128,
+            DisplaySize::Display132x64 => 64,
```
everett1992: I copied 64 from 128x32 because the display is _really_ 128x32

src/mode/terminal.rs
```diff
@@ -170,14 +170,15 @@ where
...
             DisplaySize::Display128x32 => 64,
             DisplaySize::Display96x16 => 24,
         };
 
         // Reset position so we don't end up in some random place of our cleared screen
         let (display_width, display_height) = self.properties.get_size().dimensions();
         self.properties
-            .set_draw_area((0, 0), (display_width, display_height))?;
+            .set_draw_area((6, 32), (display_width, display_height))?;
```
everett1992: These numbers are hacked in. I think I'd add a `display_start` tuple  to DisplaySize and use it anywhere set_draw_area is called to reset to the beginning of the display.

src/properties.rs
```diff
@@ -50,6 +50,7 @@ where
 
         match self.display_size {
             DisplaySize::Display128x32 => Command::ComPinConfig(false, false).send(&mut self.iface),
+            DisplaySize::Display132x64 => Command::ComPinConfig(true, false).send(&mut self.iface),
```
everett1992: I don't know what this is

src/mode/graphics.rs
jamwaffles: > These numbers look like width*height/8

Correct.

> For 132x64 I don't know if we need to draw at an offset

I don't think so; it's easier to understand if the buffer starts from zero.

> This wholeblock could be switched to this

It could, but I'd prefer to keep the explicit slice ranges. It's more obvious how many bytes are required by the display, and I want to change how this driver supports multiple resolutions soon anyway - a smaller display than the max 128x64 wastes an awful lot of RAM in the remaining part of the buffer. Tracked in #20.

src/displaysize.rs
```diff
@@ -6,6 +6,8 @@
 pub enum DisplaySize {
     /// 128 by 64 pixels
     Display128x64,
+    /// 132 by 64 pixels, try this for ssd1305 chips that are 128x64.
```
jamwaffles:
```suggestion
    /// 132 by 64 pixels. Use this for SSD1305 chips that drive 128x64px displays.
```

src/mode/terminal.rs
jamwaffles: I'm not following this comment. Shouldn't the above line be 32?

src/properties.rs
jamwaffles: Not a very well named `Command` I'll admit. The first `bool` sets whether the display lines alternate or not. If set true, the driver IC remaps the rows like this:

image

You might want to try setting the first argument to `false` to fix your "scanline" problem described in #76. I don't know what the mapping is like for SSD1305s so YMMV.

jamwaffles: The second argument defines in which direction the scan occurs. This is always false, but if your display ends up upside down (without using `Rotation`), you might want to set this to `true`.

Issue #81, opened 2019-05-19
Building examples fails due to deprecated digital::OutputPin


BenBergman:
I just tried to build the examples using a fresh clone of the repo and got a number of errors (escalated from warnings due to https://github.com/jamwaffles/ssd1306/blob/master/src/lib.rs#L160)

```
error: use of deprecated item 'hal::digital::OutputPin::set_low': Deprecated because the methods cannot return errors. Users should use the traits in digital::v2.
```

It sounds like the solution is to just update to the latest traits, so I did that in my local copy and got the examples to build. I can create a PR, but right now I'm just throwing away the result from the pin state changes (which I guess effectively matches the existing behaviour). If more thorough error handling seems appropriate I'm open to that.

jamwaffles:
Fixed in #83

PR #82, opened 2019-05-19
Update to digital v2 since original digital OutputPin has been deprecated

BenBergman:
This PR is to address issue #81


jamwaffles:
Sorry, I completely missed that this was already open when #83 was made. I've merged #83 and released 0.2.5 so you should be able to use the v2 traits with no warnings now. Thanks for the work in this PR nonetheless!

PR #83, opened 2019-06-12
Use the embedded-hal::digital::v2::OutputPin

eldruin:
Very similarly to https://github.com/jamwaffles/sh1106/pull/8, I added an error `enum` with two types: a communication error type and a pin error type which allows users to handle I2C or SPI communication errors as well as SPI pin setting errors.
I had to make `reset` fallible, though. User code will then get an unused Result warning.
No other changes are required in user code if the results were simply unwrapped.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ branch = "master"
 repository = "jamwaffles/ssd1306"
 
 [dependencies]
-embedded-hal = "0.2.2"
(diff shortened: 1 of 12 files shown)


jamwaffles:
Thanks @eldruin! I'll get this released when I'm next on a Linux machine.

jamwaffles:
Released in [0.2.5](https://crates.io/crates/ssd1306/0.2.5).

PR #86, opened 2019-07-01
Implement switching display power

### LINKED ISSUE #84: Display power off
Is it possible to power off the display using command 0xAE? This would enable saving power and OLED life in long-running applications.

tpwrules:
Implements #84 . Tested using [Adafruit OLED Featherwing](https://www.adafruit.com/product/2900). Display has no issues accepting commands while switched off so I don't believe there is a need to implement any interlocks or new state.

Closes #84

### DIFF
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -162,6 +162,11 @@ where
     pub fn set_rotation(&mut self, rot: DisplayRotation) -> Result<(), DI::Error> {
         self.properties.set_rotation(rot)
     }
+
+    /// Turn the display on or off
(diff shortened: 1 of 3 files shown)


tpwrules:
Alright, I changed it. I also added a little note clarifying when it can be used as I myself did not realize that the display could still be communicated with while turned off.

To answer your aside, I feel like it's perfectly fine for the name to start with `set_`. It's expensive in the sense that it generates bus traffic, and it changes the state of the device. I think its corresponding getter should just be `display_on` though.

jamwaffles:
Thanks for the change! I'll sanity check on my cheapo Aliexpress display module tonight and get this PR merged in.

PR #90, opened 2019-07-23
Upgrade to embedded graphics 0.6.0-alpha.2

jamwaffles:
Upgrade to embedded-graphics 0.6.0-alpha.2. This will incur an alpha release of `ssd1306` (0.3.0-alpha.1) to test e-g in the real world. The final release will need a bunch of migration guide stuff.

Cc jamwaffles/embedded-graphics#138


examples/image_i2c.rs
```diff
@@ -30,8 +30,7 @@ extern crate stm32f1xx_hal as hal;
 
 use cortex_m_rt::ExceptionFrame;
 use cortex_m_rt::{entry, exception};
-use embedded_graphics::image::Image1BPP;
-use embedded_graphics::prelude::*;
+use embedded_graphics::{image::ImageLE, prelude::*};
```
rfuest: `ImageLE` is is already included in the prelude.

examples/image_i2c.rs
```diff
@@ -75,7 +74,7 @@ fn main() -> ! {
     disp.init().unwrap();
     disp.flush().unwrap();
 
-    let im = Image1BPP::new(include_bytes!("./rust.raw"), 64, 64).translate(Point::new(32, 0));
+    let im = ImageLE::new(include_bytes!("./rust.raw"), 64, 64).translate(Point::new(32, 0));
```
rfuest: For <= 8bpp images you don't need to explicitly specify the byte order and can simply use `Image` instead.

src/mode/graphics.rs
```diff
@@ -194,19 +199,31 @@ where
...
-            self.set_pixel((pixel.0).0, (pixel.0).1, pixel.1.into_inner());
+        // Filter out pixels that are off the top left of the screen
+        let on_screen_pixels = item_pixels
+            .into_iter()
+            .filter(|drawable::Pixel(point, _)| point.x >= 0 && point.y >= 0);
+
+        for drawable::Pixel(point, color) in on_screen_pixels {
+            // NOTE: The filter above means the coordinate conversions should never panic
+            self.set_pixel(
+                point.x.try_into().expect("Point X coordinate is negative"),
```
rfuest: Using `try_into` here will probably check for `x >= 0` again, which is unnecessary and causes a slight runtime penalty.

examples/image_i2c.rs
jamwaffles: Ah nuts, I don't think any of the `Image*` exports should be in the prelude tbh. None of the primitives or fonts are so that's a mistake on my part for not spotting that in review. I'm going to open an issue in embedded-graphics to remove it.

examples/bmp_i2c.rs
```diff
@@ -0,0 +1,109 @@
...
+    // The image is an RGB565 encoded BMP, so specifying the type as `ImageBmp<Rgb565>` will read
+    // the pixels correctly
+    let im: ImageBmp<Rgb565> = ImageBmp::new(include_bytes!("./rust-pride.bmp"))
+        .expect("Failed to load BMP image")
+        .translate(Point::new(32, 0));
+
+    // The display uses `BinaryColor` pixels (on/off only). Here, we `map()` over every pixel
+    // and naively convert the colour to an on/off value. The logic below simply converts any
+    // colour that's not black into an "on" pixel.
+    let im = im.into_iter().map(|Pixel(position, colour)| {
```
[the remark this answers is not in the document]
jamwaffles: *cries in `English (GB)`*

src/mode/graphics.rs
jamwaffles: This was suggested by the compiler, but it can be simplified seeing as we do the positivity checks beforehand anyway.

examples/image_i2c.rs
jamwaffles: I get an error about being unable to resolve `BO` if I use `Image` over `ImageLE` here. I could specify the type on `im` but it's cleaner to just use `ImageLE`, no?

examples/bmp_i2c.rs
```diff
@@ -0,0 +1,109 @@
...
+    disp.flush().unwrap();
+
+    // The image is an RGB565 encoded BMP, so specifying the type as `ImageBmp<Rgb565>` will read
+    // the pixels correctly
+    let im: ImageBmp<Rgb565> = ImageBmp::new(include_bytes!("./rust-pride.bmp"))
+        .expect("Failed to load BMP image")
+        .translate(Point::new(32, 0));
+
+    // The display uses `BinaryColor` pixels (on/off only). Here, we `map()` over every pixel
+    // and naively convert the colour to an on/off value. The logic below simply converts any
```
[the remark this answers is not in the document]
jamwaffles: Changed, thanks for the suggestion :+1:

jamwaffles:
All changes made. I tested the `image_i2c` example on real hardware and it looks fine. I've resolved all comments except the two above. Thanks for the review @rfuest!

examples/image_i2c.rs
rfuest: I didn't realize that there was no type on `im`.

I prefer to explicitly set the type for images, because this specifies the data type stored in the image. If no color type is specified the compiler will infer some type, which probably matches the display, but this might not be correct for the image.

examples/image_i2c.rs
jamwaffles: For reference: #169.

examples/image_i2c.rs
jamwaffles: I've moved both image examples over to the following.

```rust
let im: Image<BinaryColor> = ...
```

Retested the `image_i2c` example and it still works (as expected).

PR #95, opened 2019-10-04
Run cargo fmt check in CI

jamwaffles:


jamwaffles:
Nightly is stalling at the moment but the rest of the build works fine, so I'mma merge anyway.

PR #106, opened 2020-02-08
Use iterators in interface send methods

jamwaffles:

## PR description

Switches some `for` loops for iterators to speed up sends (and hopefully reduce code size) by eliding bounds checks.


jamwaffles:
I just added some optimisations in 6ff90ff2abcb172db11e0d4540f8c2fcf70a4349 which should make `set_pixel` calls a lot faster. Unsure how to bench, but for one thing the change from an if/else to bit manip ops to set the pixel value in a byte should be much faster now. The [generated assembly](https://godbolt.org/z/YRSanf) says so.

jamwaffles:
Yeah, slightly bigger but maybe faster so worth it in the end? I dunno tbh, it's all tradeoffs lol. Thanks for the review!

PR #108, opened 2020-02-12
Add raw random noise example

jamwaffles:

## PR description

This is the result of an email discussion with @nebelgrau77 who is asking for assistance sending raw bytes to the display, without using `GraphicsMode`, etc. The example emulates an old TV with a white noise effect using random crap from the `rand` crate in no-std mode.

### DIFF
--- a/examples/noise_i2c.rs
+++ b/examples/noise_i2c.rs
@@ -1,0 +1,94 @@
+//! Send random raw data to the display, emulating an old untuned TV. This example `release()`s the
+//! underlying display properties struct and allows calling of the low-level `draw()` method,
+//! sending a 1024 byte buffer straight to the display.
+//!
+//! This example is for the STM32F103 "Blue Pill" board using I2C1.
+//!
+//! Wiring connections are as follows for a CRIUS-branded display:
+//!
+//! ```
+//!      Display -> Blue Pill
+//! (black)  GND -> GND
+//! (red)    +5V -> VCC
+//! (yellow) SDA -> PB9
+//! (green)  SCL -> PB8
+//! ```
+//!
+//! Run on a Blue Pill with `cargo run --example noise_i2c`. Best results when using `--release`.
+
+#![no_std]
+#![no_main]
+
+extern crate cortex_m;
+extern crate cortex_m_rt as rt;
+extern crate panic_semihosting;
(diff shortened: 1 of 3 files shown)


jamwaffles:
@therealprof This raises the question of moving some common stuff in the crate like initialisation routines into a trait so that all modes (including `RawMode` can use them). Let's leave that for another PR, but it's food for thought.

therealprof:
@jamwaffles Yes, I'm very up for that. Also we might want to reconsider the use of traits and maybe move to a simple enum instead for better code optimisation.

examples/noise_i2c.rs
```diff
@@ -0,0 +1,94 @@
...
+//!
+//! Run on a Blue Pill with `cargo run --example noise_i2c`. Best results when using `--release`.
+
+#![no_std]
+#![no_main]
+
+extern crate cortex_m;
+extern crate cortex_m_rt as rt;
+extern crate panic_semihosting;
+extern crate stm32f1xx_hal as hal;
```
therealprof: It would be great if we could slowly move to edition 2018 code. 😅

I also don't quite see the point of using semihosting in examples like this which only adds another stumbling block without any benefits.

therealprof:
Hm, I thought you re-enabled the regular merge feature? I can only squash and merge.

examples/noise_i2c.rs
jamwaffles: Yeah I want to upgrade all the examples in one big PR. Those imports are bugging me.

> I also don't quite see the point of using semihosting in examples like this

What would be a better `panic_*` to use in your opinion?

jamwaffles:
> Hm, I thought you re-enabled the regular merge feature? I can only squash and merge.

Hasn't that always been the case? At any rate, that's what I'd prefer to do anyway.

examples/noise_i2c.rs
therealprof: panic_halt or panic_abort.

therealprof:
> Hasn't that always been the case? At any rate, that's what I'd prefer to do anyway.

Rewriting the commits and autosquashing it means you're causing problems for people with work-in-progress stuff. I was hoping that the setting I recently recommended would only prevent merges of PRs with merges in them which is the bad stuff, but it seems it disables all merge commits.

PR #111, opened 2020-02-25
Add support for 64x48px SSD1306 modules

jaxter184:

- [ ] ~Check that you've added documentation to any new methods~

## PR description

I've got [a display that seems to use the SSD1306](https://www.aliexpress.com/item/32782787258.html). As far as I know, all that has to be added is a new DisplaySize enum and the corresponding match cases.

The HAL library for the MCU I have access to (STM32F303) currently has a bug that makes the SPI transfer LSBit-first, so I unfortunately can't submit an example that works out of the box. Would it be better to add an example that I can't test or to leave it to someone else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Add support for modules with a 64x48px display size.
+
 ### Changed
 
 - [#107](https://github.com/jamwaffles/ssd1306/pull/107) Migrate from Travis to CircleCI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,13 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- [#111](https://github.com/jamwaffles/ssd1306/pull/111) Fix TerminalMode offset for smaller displays.
+
 ### Added
 
-- Add support for modules with a 64x48px display size.
+- [#111](https://github.com/jamwaffles/ssd1306/pull/111) Add support for modules with a 64x48px display size.
 
 ### Changed
 
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -31,6 +31,7 @@ where
             DisplaySize::Display128x32 => (0, 0),
             DisplaySize::Display96x16 => (0, 0),
             DisplaySize::Display72x40 => (28, 0),
+            DisplaySize::Display64x48 => (32, 0),
         };
 
         DisplayProperties {
@@ -71,6 +72,7 @@ where
             DisplaySize::Display128x64 => Command::ComPinConfig(true, false).send(&mut self.iface),
             DisplaySize::Display96x16 => Command::ComPinConfig(false, false).send(&mut self.iface),
             DisplaySize::Display72x40 => Command::ComPinConfig(true, false).send(&mut self.iface),
+            DisplaySize::Display64x48 => Command::ComPinConfig(true, false).send(&mut self.iface),
         }?;
 
         Command::Contrast(0x8F).send(&mut self.iface)?;
--- a/src/displaysize.rs
+++ b/src/displaysize.rs
@@ -12,6 +12,8 @@ pub enum DisplaySize {
     Display96x16,
     /// 70 by 42 pixels
     Display72x40,
+    /// 64 by 48 pixels
+    Display64x48,
 }
 
 impl DisplaySize {
@@ -23,6 +25,7 @@ impl DisplaySize {
             DisplaySize::Display128x32 => (128, 32),
             DisplaySize::Display96x16 => (96, 16),
             DisplaySize::Display72x40 => (72, 40),
+            DisplaySize::Display64x48 => (64, 48),
         }
     }
 }
--- a/src/mode/terminal.rs
+++ b/src/mode/terminal.rs
@@ -183,6 +183,7 @@ where
             DisplaySize::Display128x32 => 64,
             DisplaySize::Display96x16 => 24,
             DisplaySize::Display72x40 => 45,
+            DisplaySize::Display64x48 => 48,
         };
 
         // Let the chip handle line wrapping so we can fill the screen with blanks faster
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Add support for modules with a 64x48px display size.
```
jamwaffles:
```suggestion
- [#111](https://github.com/jamwaffles/ssd1306/pull/111) Add support for modules with a 64x48px display size.
```

jaxter184:
> please just test that the display offsets are correct on actual hardware when you can.

The offsets seem correct from my tests. It covers the full display and (0,0) is the top left corner. I forgot to mention that I was having some issues with the terminal mode where it was line breaking before reaching the end, but I'm not sure how to fix that.

> Could you hack the 303 HAL for testing purposes?

I ended up loading [a currently pending pull request](https://github.com/stm32-rs/stm32f3xx-hal/pull/60#issue-375869909) as a local crate, so when that gets merged, I'll probably submit an example or two for the 303.

jamwaffles:
That reminds me, you need to add your display size to [`mode/terminal.rs`](https://github.com/jamwaffles/ssd1306/blob/master/src/mode/terminal.rs#L181-L186). Not sure if that'll fix your problem. Can you post a picture of the wrapping behaviour you're seeing?

jaxter184:
> That reminds me, you need to add your display size to [`mode/terminal.rs`](https://github.com/jamwaffles/ssd1306/blob/master/src/mode/terminal.rs#L181-L186).

Is there more to add other than the maximum character amount?

Here's an image of the terminal example. The diagonal lines are from the last output before this sketch, and it wraps from capital W in the next cycle. It only ever writes characters into these cells. If I run the same sketch in 64x48px on a 128x64px display, the rows extend to alternating edges of the screen.

terminal wrap

src/properties.rs
```diff
@@ -31,6 +31,7 @@ where
             DisplaySize::Display128x32 => (0, 0),
             DisplaySize::Display96x16 => (0, 0),
             DisplaySize::Display72x40 => (28, 0),
+            DisplaySize::Display64x48 => (32, 0),
```
therealprof: How did you arrive at this offset value?

jaxter184: Trial and error

therealprof: Can you try 0?

jaxter184: When it's set to 0 and the below code is run, the left half of the screen is white and the right half is black

```rust
for i in 0..64 { for j in 0..48 {
    disp.set_pixel(i,j,1);
}}
disp.flush().unwrap();
```

therealprof: It is kind of hard to follow what you're doing. Which display are we talking about now? I thought you wanted to fix the TerminalMode first?

Key to getting this fixed is the right combination of configuration values and GraphicsMode works in a totally different way to TerminalMode which is essentially setting the framebuffer values so it can automatically advance from left to right and top to bottom. If it skips to the next line after 4 characters that is typically caused by telling the display driver hardware the wrong display dimensions. It could also be that the TerminalMode calculations are wrong.

jaxter184: Ah, sorry, I didn't realize the inquiry was related to the TerminalMode issue. This is still with the 64x48px display. Changing `32` to `0` seems to result in the same output as the above image.

therealprof: Hm, so something else is wrong.

jaxter184:
I realized that at some point while converting the example, I accidentally removed the `disp.clear()` statement. After adding it back, the rows no longer alternate, but characters only show up on the left half of the screen.

therealprof:
> I realized that at some point while converting the example, I accidentally removed the `disp.clear()` statement. After adding it back, the rows no longer alternate, but characters only show up on the left half of the screen.

Which sounds correct if you're using a limited screensize on a larger display?

jaxter184:
> Which sounds correct if you're using a limited screensize on a larger display?

The half screen character output is on the 64x48px display. On the 128x64 display, it's also half of the screen (ignoring the top margin), but it has all of the letters. Based on my understanding of these displays, it seems like the terminal mode ignores the offset? Does the terminal mode work as intended on the 72x48px display?

therealprof:
> > Which sounds correct if you're using a limited screensize on a larger display?
> 
> The half screen character output is on the 64x48px display. On the 128x64 display, it's also half of the screen (ignoring the top margin), but it has all of the letters. Based on my understanding of these displays, it seems like the terminal mode ignores the offset? Does the terminal mode work as intended on the 72x48px display?

No idea really, I only have full size displays to test. Quite possible that something is missing here

jaxter184:
I've got it working on the 64x48px display by turning some `0`s into `display_x/y_offset`s, but it may have broken 72x48px displays if they used to work correctly. 128x64px displays seem to still work as they did before.

terminal fix

jaxter184:
I've squashed it down to 2 commits, should I maybe squash it down to 1?

therealprof:
As many as you'd like. Just wanted to get rid of the merge commit and the sausage making...

One more request: Would you please add a line to the ChangeLog stating that the TerminalMode for non-128x64 sizes has been fixed?

therealprof:
@jamwaffles Isn't something supposed to happen here with the CI?

jamwaffles:
Yeah I think CircleCI doesn't build pulls from forks by default. Not sure if that's the cause but it's my only theory right now. I ran the build locally and it passed, so I'm happy to merge this. 

@therealprof Do you want me to rebase and merge instead of squashing?

therealprof:
> Do you want me to rebase and merge instead of squashing?

No preference there.

I would have merged but I can't due to pending on CI.

jamwaffles:
Squashed. Let's see if master builds in CI

jamwaffles:
@jaxter184 Thanks for the PR! This will go out in the next release whenever that happens.

Sh3Rm4n:
> > Could you hack the 303 HAL for testing purposes?
> 
> I ended up loading [a currently pending pull request](https://github.com/stm32-rs/stm32f3xx-hal/pull/60#issue-375869909) as a local crate, so when that gets merged, I'll probably submit an example or two for the 303.

stm32f3xx-hal v0.4.1 is released, which includes the spi fix :)

PR #113, opened 2020-03-21
Move bounded drawing calculation from interfaces into the only call site

therealprof:
Hi! Thank you for helping out with SSD1306 development! 

Please:

## PR description

Remove some code duplication in bounded box drawing by moving the bounding box calculation from the interface specific `send_bounded_data` into the only call site and reuse the regular `send_data` function.


therealprof:
It's a bit weird that this is a world-public interface. But I've added the entry as requested.

therealprof:
NB: I've also tested the I2C version of this in one of my projects.

jamwaffles:
> It's a bit weird that this is a world-public interface. But I've added the entry as requested.

It should probably be private, no? I'm happy to make that breaking change. If someone really wants it, we can come up with a better solution down the road.

p.s. needs a rebase

therealprof:
@jamwaffles Privatizing doesn't really work, already tried. It's too exposed but hey, maybe we're completely changing it anyways. 😅

PR #116, opened 2020-03-25
Use display-interface for display communication take 2

therealprof:


Cargo.toml
```diff
@@ -19,6 +19,9 @@ targets = [ "thumbv7m-none-eabi", "thumbv7em-none-eabihf" ]
 
 [dependencies]
 embedded-hal = "0.2.3"
+display-interface = { version = "0.2" }
+display-interface-i2c = { version = "0.2" }
+display-interface-spi = { version = "0.2" }
```
jamwaffles, line 24: For consistency's sake :)

```suggestion
display-interface = "0.2"
display-interface-i2c = "0.2"
display-interface-spi = "0.2"
```

src/builder.rs
```diff
@@ -2,11 +2,12 @@
...
-//! `.connect_spi(spi, dc)`. The builder will be consumed into a
-//! [`mode::RawMode`](../mode/raw/struct.RawMode.html) object which can be coerced into a richer
-//! display mode like [mode::Graphics](../mode/graphics/struct.GraphicsMode.html) for drawing
-//! primitives and text.
+//! To finish the builder and produce a connected display interface, call `.connect(interface)`
+//! where `interface` is an instantiated `DisplayInterface` implementation . For I2C interfaces
+//! there's also a `I2CDIBuilder` to simplify the construction of a I2C DisplayInterface. The
+//! builder will be consumed into a [`mode::RawMode`](../mode/raw/struct.RawMode.html) object
+//! which can be coerced into a richer display mode like
+//! [mode::Graphics](../mode/graphics/struct.GraphicsMode.html) for drawing primitives and text.
```
jamwaffles, line 10: I've peppered a few formatting and wording fixes throughout this para. Github's suggestion syntax doesn't show the character-level changes unfortunately :/

```suggestion
//! To finish the builder and produce a connected display interface, call `.connect(interface)`
//! where `interface` is an instantiated `DisplayInterface` implementation. For I2C interfaces
//! there's also an [`I2CDIBuilder`] to simplify the construction of an I2C `DisplayInterface`. The
//! builder will be consumed into a [`mode::RawMode`] object which can be coerced into a richer 
//! display mode like [`mode::Graphics`].
//!
//! [`I2CDIBuilder`]: ./struct.I2CDIBuilder.html
//! [`mode::RawMode`]: ../mode/raw/struct.RawMode.html
//! [`mode::Graphics`]: ../mode/graphics/struct.GraphicsMode.html
```

Please check the doc links in my edit - I'm not sure they're all correct.

src/builder.rs
```diff
@@ -90,49 +90,53 @@ impl Builder {
...
+}
+
+impl I2CDIBuilder {
+    /// Create new builder with a default I2C address of 0x3C
+    pub fn new() -> Self {
+        Self { i2c_addr: 0x3c }
+    }
+
+    /// Set the I2C address to use. Defaults to 0x3C which is the most common address.
+    /// The other address specified in the datasheet is 0x3D
```
jamwaffles, line 131:
```suggestion
    /// Set the I2C address to use
    ///
    /// [`I2CDIBuilder::new`] defaults to an address of `0x3C` which is the most common address.
    /// The other address specified in the datasheet is `0x3D` which can be set using this method.
```

src/builder.rs
```diff
@@ -90,49 +90,53 @@ impl Builder {
...
-        DisplayMode::<RawMode<I2cInterface<I2C>>>::new(properties)
+        let properties = DisplayProperties::new(interface, self.display_size, self.rotation);
+        DisplayMode::<RawMode<I>>::new(properties)
     }
+}
 
-    /// Finish the builder and use SPI to communicate with the display
+/// Builder struct for an I2C interface. Driver options and interface are set using its methods.
+#[derive(Clone, Copy)]
+pub struct I2CDIBuilder {
```
rfuest, line 114: Shouldn't this be called `I2cDiBuilder` to be compliant with the rust naming conventions? I think that `I2c` is also more common in HALs than `I2C`.

therealprof: Well, I2C is the real acronym, just like SPI. I don't really have any preference in how to call it but it's somewhat unfortunate to use an incorrectly spelled just for programming language naming convention sake.

If you prefer me to make that change, then no biggie.

rfuest: > Well, I2C is the real acronym, just like SPI. I don't really have any preference in how to call it but it's somewhat unfortunate to use an incorrectly spelled just for programming language naming convention sake.

IMO the reason to spell it incorrectly is increased readability, which was the primary reason I brought this up. `I2CDI` is harder to decipher than `I2cDi`. And this also makes this code more consistency with other crates in the embedded rust ecosystem, like the spelling of `Pwm` and `Qei` in `embedded_hal`.

For this reasons I would prefer to use `I2c` and `Spi` in this crate, but also in `display-interface`.

therealprof: Yeah, I'm definitely not going to change `display-interface`. I don't like the spelling of I2c or Spi at all (neither do I like Pwm and Qei, but they are what they are for whatever reason and are not going to change). If @jamwaffles would prefer `I2cDi` or maybe `I2CDi` I'd be happy to change it.

jamwaffles: I have opinions both ways lol. Both versions read strangely to me due to the pile of acronyms in the names. `display_interface_i2c` uses `I2CInterface` casing, so let's keep this PR as `I2CDIBuilder` for some semblance of consistency with that crate.

At the end of the day, these identifiers are going to be copypasta-ed and forgotten about so let's call this bikeshed resolved :).

PR #121, opened 2020-05-28
Brightness change

bugadani:

## PR description

Some open questions that I still have:
 - Should there be an option for custom brightness levels? Adjusting the precharge period can make the screen go darker than default, but the results are not straight forward
 - Should there be an additional method per display mode? I personally prefer the `change_property` with a closure, because it allows for the same (and more) functionality without implementing a bunch of forwarder methods. Or maybe it is overly generic in the current form.

So for now, this is only a draft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Added
 
-- [#121](https://github.com/jamwaffles/ssd1306/pull/121) Brightness control
+- [#121](https://github.com/jamwaffles/ssd1306/pull/121) Added brightness control with the `set_brightness()` method.
 - [#118](https://github.com/jamwaffles/ssd1306/pull/118) `DisplayModeTrait::into_properties()` new method that consumes the driver and returns the `DisplayProperties`
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Added
 
+- [#121](https://github.com/jamwaffles/ssd1306/pull/121) Brightness control
 - [#118](https://github.com/jamwaffles/ssd1306/pull/118) `DisplayModeTrait::into_properties()` new method that consumes the driver and returns the `DisplayProperties`
 
 ### Changed
--- a/examples/rtfm_brightness.rs
+++ b/examples/rtfm_brightness.rs
@@ -1,4 +1,4 @@
-//! Cycle the screen brightness through 5 predefined brightness levels when the "DVD" logo hits one 
+//! Cycle the screen brightness through 5 predefined brightness levels when the "DVD" logo hits one
 //! of the sides of the display.
 //!
 //! For best results, run with the `--release` flag.
--- a/examples/rtfm_brightness.rs
+++ b/examples/rtfm_brightness.rs
@@ -169,7 +169,7 @@ const APP: () = {
                 };
 
                 // Send the new brightness value to the display
-                display.change_brightness(*brightness).unwrap();
+                display.set_brightness(*brightness).unwrap();
             }
         }
 
--- a/examples/rtfm_brightness.rs
+++ b/examples/rtfm_brightness.rs
@@ -1,4 +1,5 @@
-//! Same as the rtfm_brightness example, but when the logo hits one of the sides, change brightness.
+//! Cycle the screen brightness through 5 predefined brightness levels when the "DVD" logo hits one 
+//! of the sides of the display.
 //!
 //! For best results, run with the `--release` flag.
 
--- a/examples/rtfm_brightness.rs
+++ b/examples/rtfm_brightness.rs
@@ -1,0 +1,191 @@
+//! Same as the rtfm_brightness example, but when the logo hits one of the sides, change brightness.
+//!
+//! For best results, run with the `--release` flag.
+
+#![no_std]
+#![no_main]
+
+use core::convert::TryFrom;
+use display_interface_spi::SPIInterfaceNoCS;
+use embedded_graphics::{
+    geometry::Point, image::Image, pixelcolor::BinaryColor, prelude::*, primitives::Rectangle,
+};
+use panic_halt as _;
+use rtfm::app;
+use ssd1306::{prelude::*, Builder};
+use stm32f1xx_hal::{
+    delay::Delay,
+    gpio,
+    pac::{self, SPI1},
+    prelude::*,
+    spi::{self, Mode, Phase, Polarity, Spi},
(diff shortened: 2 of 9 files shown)


src/brightness.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// A bright predefined brightness level
+    pub const BRIGHT: Brightness    = Brightness::custom(0x2, 0x9F);
+
+    /// The brightest predefined brightness level
+    pub const BRIGHTEST: Brightness = Brightness::custom(0x2, 0xFF);
+
+    /// Create a Brightness object from a precharge period and contrast pair.
+    ///
+    /// `precharge` must be between 1 and 15
```
jamwaffles, line 28: There's no mention of the acceptable values for `contrast` here. It might also be useful to state the appropriate sections of the datasheet. For example:

```suggestion
    /// Create a Brightness object from a precharge period and contrast pair.
    ///
    /// `precharge` sets the `phase 1` argument of the `0xD9 Set Pre-Charge Period` command and must 
    /// be must be between 1 and 15. See section 10.1.17 of the SSD1306 datasheet for more 
    /// information.
    ///
    /// `contrast` sets the value used in the `0x81 Set Contrast Control` command and must be 
    /// between 0 and 255. See section 10.1.7 of the SSD1306 datasheet for more information.
```

src/brightness.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Struct that holds display brightness
+#[derive(Copy, Clone)]
+pub struct Brightness {
+    pub(crate) precharge: u8,
+    pub(crate) contrast: u8,
+}
+
+impl Brightness {
+    /// The darkest predefined brightness level
+    pub const DARKEST: Brightness   = Brightness::custom(0x1, 0x00);
```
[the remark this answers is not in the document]
bugadani: Haha, fair enough

src/brightness.rs
bugadani: Technically you are right, and a reference certainly doesn't hurt, but I found the values behave rather ... inconsistently. Technically both precharge periods, the Vcomh deselect value and the contrast affect the brightness, but my results were a hot mess, that's why only 2 of those are included in this PR. Probably needs some more thought.

bugadani:
I believe the various display properties need some thought here. I prefer the closure method for setting them, because it doesn't require implementing the same functionality in like 3 places ( one impl and 2 forwarders in the display modes), but your request is totally sensible. Should I include other properties in the common set (i.e. the trait)?

bugadani:
I've copied the rtfm_dvd example and set it up so when the logo hits the sides, the brightness changes. Unfortunately, I have no hardware to test it.

bugadani:
I really don't have a better idea of how to implement this in a somewhat nice manner, so I'm marking the PR as ready.

src/mode/terminal.rs
```diff
@@ -278,6 +279,11 @@ where
         self.properties.display_on(on).terminal_err()
     }
 
+    /// Change the display brightness.
+    pub fn change_brightness(&mut self, brightness: Brightness) -> Result<(), TerminalModeError> {
```
jamwaffles, line 283: The other setters are prefixed with `set_`, not `changed_` so for consistency:

```suggestion
    pub fn set_brightness(&mut self, brightness: Brightness) -> Result<(), TerminalModeError> {
```

src/properties.rs
```diff
@@ -251,6 +251,18 @@ where
         Command::DisplayOn(on).send(&mut self.iface)
     }
 
+    /// Change the display brightness.
+    pub fn change_brightness(&mut self, brightness: Brightness) -> Result<(), DisplayError> {
```
jamwaffles, line 255: Likewise here

```suggestion
    pub fn set_brightness(&mut self, brightness: Brightness) -> Result<(), DisplayError> {
```

examples/rtfm_brightness.rs
```diff
@@ -0,0 +1,191 @@
+//! Same as the rtfm_brightness example, but when the logo hits one of the sides, change brightness.
```
jamwaffles, line 1:
```suggestion
//! Cycle the screen brightness through 5 predefined brightness levels when the "DVD" logo hits one 
//! of the sides of the display.
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Added
 
+- [#121](https://github.com/jamwaffles/ssd1306/pull/121) Brightness control
```
jamwaffles, line 11:
```suggestion
- [#121](https://github.com/jamwaffles/ssd1306/pull/121) Added brightness control with the `set_brightness()` method.
```

`set_brightness` here assuming you've switched `change_brightness` to `set_brightness` in the code.

src/command.rs
```diff
@@ -76,7 +76,7 @@ pub enum Command {
     /// First value is oscillator frequency, increasing with higher value
     /// Second value is divide ratio - 1
     DisplayClockDiv(u8, u8),
-    /// Set up phase 1 and 2 of precharge period. each value is from 0-63
+    /// Set up phase 1 and 2 of precharge period. each value is from 1-15
```
jamwaffles, line 79:
```suggestion
    /// Set up phase 1 and 2 of precharge period. Each value must be in the range 1 - 15.
```

src/mode/terminal.rs
bugadani: I'm usually against the get/set wording, as explained by Kevlin Henney, those are the most overloaded words in the English language. I know it's against the tradition, but that's the reason if you ever catch me not naming something `set` :)

jamwaffles: Agreed. I wrote these methods before I knew about the [Rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter). There's nothing in there about setters, but we could certainly remove the `get_` prefixes in a future PR to bring this crate in line with the guidelines.

src/command.rs
bugadani: At least that wording wasn't me :P

jamwaffles:
I just merged #126 so if you can give this a rebase I'll approve and merge once the build passes.

jamwaffles:
The build is failing with some unused/not found errors.

Once fixed, you can try running `TARGET=thumbv7m-none-eabi ./build.sh` on your PC to test things before pushing if you want.

bugadani:
Ah yes, I renamed change_brightness to set, but forgot the example... Oh well

bugadani:
That's better

jamwaffles:
Thanks!

PR #123, opened 2020-06-07
Clean up graphics mode

### LINKED ISSUE #122: Incorrect min_y?
Based on [this metric](http://reviewthecode.blogspot.com/2016/01/wtf-per-minute-actual-measurement-for.html) I believe the linked line may be incorrect.

https://github.com/jamwaffles/ssd1306/blob/96170fd21419e73fdb2299f395b07836d52d090f/src/mode/graphics.rs#L161

If that's supposed to be height, I'm happy to fix it, otherwise please provide an explanation why it's width.

bugadani:
Extract resetting of the dirty area - use constants, since the actual values are irrelevant, only their relation is important for flush to return early

Remove redundant checks and min/max calls - assuming width/height is always divisible by 8, set_pixel will make sure max_x and max_y is in the valid range

~- [ ] Add or modify an example if there are changes to the public API~
~- [ ] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, **Changed**, etc)~ No visible changes

## PR description

This needs some double checking because my understanding may be incorrect.

Closes #122 but that is not a bug; may result in a negligible amount of perf gain (some checks per frame).

### DIFF
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -96,6 +96,16 @@ impl<DI> GraphicsMode<DI>
 where
     DI: WriteOnlyDataCommand,
 {
+    /// Resets the area where the framebuffer was modified.
+    fn reset_dirty_area(&mut self) {
+        //  min > max means no set_pixel calls were made so far
+        self.min_x = 255;
+        self.max_x = 0;
+
+        self.min_y = 255;
+        self.max_y = 0;
+    }
+
     /// Clear the display buffer. You need to call `disp.flush()` for any effect on the screen
     pub fn clear(&mut self) {
         self.buffer = [0; 1024];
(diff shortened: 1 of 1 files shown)


jamwaffles:
I haven't got to the bottom of exactly why yet, but the `rtfm_dvd` example crashes when the image touches the right hand side of the display. This doesn't occur in master so I'm assuming there's an overflow error introduced in this branch somewhere.

bugadani:
Well, that means my logic was incorrect somewhere. Thanks, I'll check that. In the mean time, this gets demoted to draft :)

bugadani:
Printing the string "Release to shut down" before took 128348 cycles, with this patch 124478 cycles, around 3% improvement. Which doesn't mean a lot if it crashes. Investigating.

bugadani:
The collision detection in `rtfm_dvd` triggers when the right edge of the image is outside of the display area, and that crashes because max_x (or max_y) can get larger than the width/height of the screen. This is because I removed the max calls in `flush()` and set_pixel does not validate the coordinates - this also means the rendered image could be incorrect in that case, even if I reintroduce the `max` calls.

bugadani:
I've reintroduced clipping the display area to its physical size, but the possibility of incorrect drawing is still there in `set_pixel`.

bugadani:
I'm closing this in favour of #125

PR #124, opened 2020-06-08
Different GraphicsMode optimization options

bugadani:

- [ ] Rebase from `master` if you're not already up to date
- [ ] Run `rustfmt` on the project with `cargo fmt --all` - CI will not pass without this step
- [ ] Check that your branch is up to date with master and that CI is passing once the PR is opened

## PR description

In my use case, I'm using the display on an 8MHz SPI connection, which means a flush is around 1ms. This allows me to switch the partial update to a full update, and basically eliminate tracking of the dirty area.

I still need to benchmark this to see if there is any point in adding the extra complexity, but I wanted to open this draft to hear some of your thoughts on the matter.

By default, the implementation uses the old dirty area tracking method with the partial screen update, so it's hopefully backwards compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Added
 
+- [#124](https://github.com/jamwaffles/ssd1306/pull/124) Added type parameter to optimize `GraphicsMode` for draw speed instead of flush time.
 - [#121](https://github.com/jamwaffles/ssd1306/pull/121) Added brightness control with the `set_brightness()` method.
 - [#118](https://github.com/jamwaffles/ssd1306/pull/118) `DisplayModeTrait::into_properties()` new method that consumes the driver and returns the `DisplayProperties`
 
--- a/src/mode/graphics.rs
+++ b/src/mode/graphics.rs
@@ -62,27 +62,145 @@ use crate::{
     properties::DisplayProperties,
 };
 
-// TODO: Add to prelude
-/// Graphics mode handler
-pub struct GraphicsMode<DI> {
-    properties: DisplayProperties<DI>,
-    buffer: [u8; 1024],
+#[derive(Copy, Clone)]
+/// This optimization method keeps track of the area that needs to be updated,
+/// resulting in a potentially faster flush process.
+pub struct DirtyArea {
     min_x: u8,
     max_x: u8,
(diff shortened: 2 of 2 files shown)


bugadani:
Oh, also this is based on #123, hence the duplicate commits.

bugadani:
Just in case, please don't waste time on reviewing the code yet (since this was thrown together extremely quickly and has a ton of potential issues, including the crash of #123), only the general idea :)

bugadani:
A speed bump of around 8% (117813 cycles instead of 128348 while drawing a piece of text) when using FastSetPixel. Tested on a nRF52832 @ 64MHz

bugadani:
Since it's possible to draw to an otherwise unexpected location, I'm thinking about making it possible in FastSetPixel (or a different UncheckedFastSetPixel) to write to an unchecked address. It's already up to the user to draw to the right area, so I think an explicitly unsafe option is not that big deal to add.

bugadani:
Closes #122

bugadani:
I'm done modifying this. There is a single compare that could be optimized away iff it can be ensured that every write is inside the display area, but I don't think it would be worth the extra complexity and the risk.

jamwaffles:
I like the idea of moving the dirty tracker to a separate struct for cleanliness, as well as the fix for #122, but I'm not sure I see a benefit to making it customisable with another associated type, particularly if we add this one on top of those added in #125. Is 1ms update really too slow for your application?

bugadani:
My benchmark is my original C code that was able to display a realtime sampled signal at 100fps. While I don't expect the same performance, I'm not even close to that with this patch either :) After #125, most of the set_pixel code is the 4 min/max calls, so yeah, I would like to see those go away.

therealprof:
100fps? I take it that's not over 400kHz I2C... It seems dubious at least that the min/max would significantly slow down the drawing.

bugadani:
I'm using 8MHz SPI

bugadani:
I don't know why I didn't think of this, but I could just use a tailor-made graphics mode implementation for my use case...

bugadani:
I'm closing this, if someone wants their `GraphicsMode` to be faster, they can reimplement it easy enough and that basically makes the added complexity of this PR not worth breaking everybody's code.

PR #125, opened 2020-06-11
Type-level display size

### LINKED ISSUE #20: More efficient display buffer

### LINKED ISSUE #122: Incorrect min_y?

bugadani:

- [ ] Rebase from `master` if you're not already up to date
- [ ] Run `rustfmt` on the project with `cargo fmt --all` - CI will not pass without this step
- [ ] Check that your branch is up to date with master and that CI is passing once the PR is opened

## PR description

This is just a rough draft of how I imagine type-level display sizes. This has several added benefits:
 - allows the user to specify display sizes
- ~can be extended via GenericArray to support flexible display buffer size~ accidentally implemented
 - hopefully allows generating much nicer machine code

The downside is the amount of code changed and the API it broke...

Closes #20 
Closes #122

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ embedded-hal = "0.2.3"
 display-interface = "0.4"
 display-interface-i2c = "0.4"
 display-interface-spi = "0.4"
+typenum = "1.12.0"
 
 [dependencies.embedded-graphics]
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ display-interface = "0.4"
 display-interface-i2c = "0.4"
 display-interface-spi = "0.4"
 typenum = "1.12.0"
+generic-array = "0.14.2"
 
 [dependencies.embedded-graphics]
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ embedded-hal = "0.2.3"
 display-interface = "0.4"
 display-interface-i2c = "0.4"
 display-interface-spi = "0.4"
-typenum = "1.12.0"
 generic-array = "0.14.2"
 
 [dependencies.embedded-graphics]
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -58,15 +58,13 @@
 
 use display_interface::WriteOnlyDataCommand;
 
-use crate::{
-    displayrotation::DisplayRotation, displaysize::*, properties::DisplayProperties,
-};
+use crate::{displayrotation::DisplayRotation, displaysize::*, properties::DisplayProperties};
 
 /// Builder struct. Driver options and interface are set using its methods.
 #[derive(Clone, Copy)]
-pub struct Builder<DSIZE=DisplaySize128x64>
+pub struct Builder<DSIZE = DisplaySize128x64>
 where
-    DSIZE: DisplaySize
+    DSIZE: DisplaySize,
 {
     rotation: DisplayRotation,
     _size: core::marker::PhantomData<DSIZE>,
@@ -83,7 +81,7 @@ where
 
 impl<DSIZE> Builder<DSIZE>
 where
-    DSIZE: DisplaySize
+    DSIZE: DisplaySize,
 {
     /// Create new builder with a default size of 128 x 64 pixels and no rotation.
     pub fn new() -> Self {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -33,7 +33,7 @@
 //! let interface = I2CDIBuilder::new().init(i2c);
 //! Builder::new()
 //!     .with_rotation(DisplayRotation::Rotate180)
-//!     .size(DisplaySize::Display128x32)
+//!     .size::<DisplaySize128x32>()
 //!     .connect(interface);
 //! ```
 //!
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -70,19 +70,13 @@ where
     _size: core::marker::PhantomData<DSIZE>,
 }
 
-impl<DSIZE> Default for Builder<DSIZE>
-where
-    DSIZE: DisplaySize,
-{
+impl Default for Builder {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<DSIZE> Builder<DSIZE>
-where
-    DSIZE: DisplaySize,
-{
+impl Builder {
     /// Create new builder with a default size of 128 x 64 pixels and no rotation.
     pub fn new() -> Self {
         Self {
@@ -90,6 +84,12 @@ where
             _size: core::marker::PhantomData,
         }
     }
+}
+
+impl<DSIZE> Builder<DSIZE>
+where
+    DSIZE: DisplaySize,
+{
 
     /// Set the size of the display. Supported sizes are defined by [DisplaySize].
     pub fn size<SIZE: DisplaySize>(self) -> Builder<SIZE> {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -33,7 +33,7 @@
 //! let interface = I2CDIBuilder::new().init(i2c);
 //! Builder::new()
 //!     .with_rotation(DisplayRotation::Rotate180)
-//!     .size::<DisplaySize128x32>()
+//!     .size(DisplaySize128x32)
 //!     .connect(interface);
 //! ```
 //!
@@ -66,8 +66,8 @@ pub struct Builder<DSIZE = DisplaySize128x64>
 where
     DSIZE: DisplaySize,
 {
+    size: DSIZE,
     rotation: DisplayRotation,
-    _size: core::marker::PhantomData<DSIZE>,
 }
 
 impl Default for Builder {
@@ -80,8 +80,8 @@ impl Builder {
     /// Create new builder with a default size of 128 x 64 pixels and no rotation.
     pub fn new() -> Self {
         Self {
+            size: DisplaySize128x64,
             rotation: DisplayRotation::Rotate0,
-            _size: core::marker::PhantomData,
         }
     }
 }
@@ -91,10 +91,10 @@ where
     DSIZE: DisplaySize,
 {
     /// Set the size of the display. Supported sizes are defined by [DisplaySize].
-    pub fn size<SIZE: DisplaySize>(self) -> Builder<SIZE> {
+    pub fn size<S: DisplaySize>(self, size: S) -> Builder<S> {
         Builder {
+            size,
             rotation: self.rotation,
-            _size: core::marker::PhantomData,
         }
     }
 
@@ -112,7 +112,7 @@ where
     where
         I: WriteOnlyDataCommand,
     {
-        DisplayProperties::new(interface, self.rotation)
+        DisplayProperties::new(interface, self.size, self.rotation)
     }
 }
 
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -31,7 +31,21 @@
 //! use ssd1306::{prelude::*, Builder, I2CDIBuilder};
 //!
 //! let interface = I2CDIBuilder::new().init(i2c);
-//! Builder::new()
+//! let di: DisplayProperties<_> = Builder::new()
+//!     .with_rotation(DisplayRotation::Rotate180)
+//!     .connect(interface);
+//! ```
+//!
+//! When using a display other than the 128 x 64, you need to specify the display size in the
+//! second type parameter:
+//!
+//! ```rust
+//! # use ssd1306::test_helpers::{PinStub, I2cStub};
+//! # let i2c = I2cStub;
+//! use ssd1306::{prelude::*, Builder, I2CDIBuilder};
+//!
+//! let interface = I2CDIBuilder::new().init(i2c);
+//! let di: DisplayProperties<_, _> = Builder::new()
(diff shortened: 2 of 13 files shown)


bugadani:
I wish there was a way to skip CI on drafts... this really is a waste of energy at this point

bugadani:
Haven't measured the perf difference yet, but setting no_inline and inspecting the generated assembly, at least there are no integer multiplications generated. Which is nice IMO :)

src/displaysize.rs
```diff
@@ -1,31 +1,82 @@
...
+
+pub trait DisplaySize {
+    type Width: Unsigned;
+    type Height: Unsigned;
+    type OffsetX: Unsigned;
+    type OffsetY: Unsigned;
+    type BufferSize: ArrayLength<u8>;
+
+    fn ComPinConfig() -> Command;
+}
```
jamwaffles, line 17: This trait would be better implemented as:

```rust
pub trait DisplaySize {
    const WIDTH: u8;
    const HEIGHT: u8;
    const OFFSETX: u8 = 0;
    const OFFSETY: u8 = 0;
    type BufferSize: ArrayLength<u8>;

    fn com_pin_config() -> Command;
}
```

1. Associated types are moved to constants
2. Display sizes that don't require an offset can omit them from their `DisplaySize` impls.
3. `ComPinConfig` should be changed to `com_pin_config` to better match Rust API guidelines

src/displaysize.rs
```diff
@@ -1,31 +1,82 @@
...
+
+pub struct DisplaySize128x64;
+impl DisplaySize for DisplaySize128x64 {
+    type Width = U128;
+    type Height = U64;
+    type OffsetX = U0;
+    type OffsetY = U0;
```
jamwaffles, line 24: With `const` changes suggested above:

```suggestion
    const WIDTH: u8 = 128;
    const HEIGHT: u8 = 64;
```

src/mode/terminal.rs
```diff
@@ -41,6 +42,30 @@ use crate::{
 use core::{cmp::min, fmt};
 use hal::{blocking::delay::DelayMs, digital::v2::OutputPin};
 
+pub trait TerminalDisplaySize: DisplaySize {
+    type CharNum: Unsigned;
```
jamwaffles, line 46: Similar to my `DisplaySize` comment, this could be:

```suggestion
    const CHARNUM: u8;
```

src/displaysize.rs
bugadani: Thanks, good suggestions

bugadani:
Indeed, the associated constants are nice. I guess I went trigger-happy with the typenum crate

bugadani:
Now that's pedantic, failing build because of an extra empty line 😂

bugadani:
I'm quite sure the display rotation is incorrect. If I remembed my coordinate geometry, a rotate is a switch of the coordinates and the mirroring of one of them, for example in this case (x, y) should be mapped to (H-y, x) where H is the display height. I'm afraid including that will become somewhat messy, but we'll see :)

bugadani:
Currently, changing the rotation is only possible via a complete reconfiguration. I'm not sure if I want to reintroduce it, though, since the type would change and the interface errors should also be returned, so a function like that would have a horrible signature.

bugadani:
```asm
	push	{r4, r5, r6, r7, lr}
	add	r7, sp, #12
	str	r8, [sp, #-4]!
	and	r12, r2, #248
	mov.w	lr, #0
	lsl.w	r12, r12, #4
	uxtab	r12, r12, r1
	cmp.w	lr, r12, lsr #10
	bne	.LBB35_2
	ldrb.w	lr, [r0, #1044]
	uxtb	r1, r1
	ldrb.w	r5, [r0, #1046]
	uxtb	r4, r2
	ldrb.w	r8, [r0, #1045]
	cmp	lr, r1
	ldrb.w	r6, [r0, #1047]
	it	hi
	movhi	lr, r1
	cmp	r5, r4
	add	r12, r0
	strb.w	lr, [r0, #1044]
	it	hi
	movhi	r5, r2
	cmp	r8, r1
	strb.w	r5, [r0, #1046]
	it	hi
	movhi	r1, r8
	add.w	r12, r12, #20
	strb.w	r1, [r0, #1045]
	cmp	r6, r4
	it	ls
	movls	r6, r2
	strb.w	r6, [r0, #1047]
	and	r0, r2, #7
	ldrb.w	r1, [r12]
	lsl.w	r2, r3, r0
	movs	r3, #1
	lsl.w	r0, r3, r0
	bic.w	r0, r1, r0
	orrs	r0, r2
	strb.w	r0, [r12]
	ldr	r8, [sp], #4
	pop	{r4, r5, r6, r7, pc}
```

Not bad. The 5 conditionals are (well, should be, haven't really checked that close) the index check and the 4 min/max calls. Funny that half of the assembly is related to the min/max tracking :)

therealprof:
> Currently, changing the rotation is only possible via a complete reconfiguration. I'm not sure if I want to reintroduce it, though, since the type would change and the interface errors should also be returned, so a function like that would have a horrible signature.

Shouldn't that be rather easy to implement? I mean, it's basically just a function telling the display to interpret the pixels differently and then change the typestate into the new rotational typestate. The HAL impls are doing the very same all the time for e.g. the GPIO pins.

bugadani:
Yes and no. Handling failure can get messy, because the func must return the old object and the error. In fact, if one if the two commands fail for some reason, there may not be a clear recovery path. I may be overcomplicating things but I don't necessarily want a fire and forget configuration.

therealprof:
If your I2C communication fails during configuration it's game over anyway and you can only start over. There's no way of telling where it failed... Thinking about it some more it probably doesn't make too much sense to offer live rotation anyway since it's way too complex too handle, just changing the display setup is not enough, you'd actually have to repaint the whole screen for which you need new pixel data in case you're rotating by 90° or 270° since the aspect ratio most likely changes.

bugadani:
I really don't think trying to figure out error handling is in the scope of a display driver. However, I believe consuming a display interface in case of an error is impolite, panicing is simply wrong and eating the error is also incorrect.

I mostly agree with "If your I2C communication fails during configuration it's game over anyway", but theoretically it's not always a communication failure. What if someone implements a virtual driver that can fail to take the peripherial and returns that as an error? Then there is the possiblity to retry later.

I'm also worried if easy change of typestate rotation would possibly cause binary bloat, since every draw operation would be duplicated for the rotation types (the whole point of this PR is an optimized draw operation support), and I'm fairly certain rustc isn't smart enough to deduplicate.

There's also the ease of use aspect: if I just want to rotate my screen, do I really need to add an awful lot of boilerplate code to support plugging in a rotation value into my display driver?

I've implemented dynamic rotation to allow a tradeoff between convenience and performance. However, there are still several open topics in this PR:

- [ ] Names (I could certainly use some suggestions here)
- [ ] What should be the default rotation (I see arguments for both fixed 0 and dynamic)
- [ ] Formatting
- [ ] Documentation
- [ ] Getting rid of * imports
- [ ] Examples

therealprof:
> I mostly agree with "If your I2C communication fails during configuration it's game over anyway", but theoretically it's not always a communication failure. What if someone implements a virtual driver that can fail to take the peripherial and returns that as an error? Then there is the possiblity to retry later.

Maybe, maybe not. The problem is that there's a sequence of steps which need to be followed to change the display configuration and if any error occurs the display really could be in any state (including lock-up) and neither the user nor the driver know which one that is. The only sane way to handle this IMHO would be to give the resources back and let the application decide what to do with them.

therealprof:
NB: Those comments are just some food for thought. I don't care either way whether and how we support screen rotation at runtime because I'm certain it's not easy to cover every of the many angles and I'm also not conviced there's great benefit to supporting that. It'd be much easier to allow for proper display destruction and have an application needing that feature reinitialize the display, that way the behaviour is clear.

bugadani:
> NB: Those comments are just some food for thought. I don't care either way whether and how we support screen rotation at runtime because I'm certain it's not easy to cover every of the many angles and I'm also not conviced there's great benefit to supporting that. It'd be much easier to allow for proper display destruction and have an application needing that feature reinitialize the display, that way the behaviour is clear.

I'd like to add that we're in no rush to get this merged, if it ever will be :) I'm open to suggestions if the current solution is not satisfactory for any reason. In fact, I'm quite aware that having 3 (and with #124, 4) template parameters are rather overwhelming to work with, but my main goal was to get this working first and clean up the specifics later if necessary.

bugadani:
Well, that looks a lot nicer, unit-like tuple structs are quite useful.

Do you have opinions about changing rotation to orientation? I believe it's a bit more straightforward, since display orientation is actually a display property, while image rotation is kinda related to the render part?
I find myself often confused by what a 90° image rotation means, on an intuitive level it feels backwards for me. "I rotate the image 90° clockwise, which means I need to rotate my device 90° CCW".

therealprof:
I don't have any opinion on the naming. Either seems fine to me.

jamwaffles:
The type-level display _size_ makes sense to me as it can never change, but why type-level rotation? Is it for optimisation reasons? The distinction between fixed and non-fixed rotations is odd as well. 

> I find myself often confused by what a 90° image rotation means, on an intuitive level it feels backwards for me. "I rotate the image 90° clockwise, which means I need to rotate my device 90° CCW".

For this example, imagine a device where the display is screwed into the case rotated 90deg CCW. In this case, the image would need to be rotate 90deg CW to show in the correct orientation to the user. We could reduce confusion by rename `DisplayRotation` to `ImageRotation`. I'd prefer to stick with "rotation" over "orientation" unless there are objections.

bugadani:
Yes, the rotation is there for optimization. Fixed produces the simplest machine code, but takes away `set_rotation`, which is problematic to implement with changing typestate, and can also be done by reinitializing the display anyway. The non-fixed rotation adds the `set_rotation` method, and it is used by default to keep compatibility. I'm sure the wording can be improved somehow.

Renaming to ImageRotation seems like a good compromise to ease my confusion :)

bugadani:
For now I've removed the rotation related changes from this PR.

The current PR contains the following:
 - remove the old DisplaySize enum
 - add new unit-like display size types
 - adds a type parameter to *Mode and DisplayProperties, the default is the 128x64 display
 - source compatibility for anyone who used the default display size
 - only the necessary size framebuffer is allocated
 - Commands are made public, in order to allow custom display sizes
 - displaysize module is public, but reexported via prelude (umm... yeah, this is probably not nice)

src/builder.rs
```diff
@@ -31,9 +31,23 @@
...
 //! let interface = I2CDIBuilder::new().init(i2c);
-//! Builder::new()
+//! let di: DisplayProperties<_> = Builder::new()
 //!     .with_rotation(DisplayRotation::Rotate180)
-//!     .size(DisplaySize::Display128x32)
+//!     .connect(interface);
+//! ```
+//!
+//! When using a display other than the 128 x 64, you need to specify the display size in the
+//! second type parameter:
```
jamwaffles, line 40:
```suggestion
//! The builder defaults to a display size of 128 x 64px. To use a display with a different size,
//! call the [`size`](#method.size) method. Supported sizes can be found in the
//! [`displaysize`](../displaysize/index.html) module or in the [prelude](../prelude/index.html).
```

src/displaysize.rs
```diff
@@ -1,31 +1,101 @@
...
-    /// 128 by 32 pixels
-    Display128x32,
-    /// 96 by 16 pixels
-    Display96x16,
-    /// 70 by 42 pixels
-    Display72x40,
-    /// 64 by 48 pixels
-    Display64x48,
+// No need to implement Copy
+#![allow(missing_copy_implementations)]
```
jamwaffles, line 4: I agree it's a bit superfluous, but it can't hurt to add `#[derive(Debug, Copy, Clone)]` to all the size structs IMO. Someone out there might find a use for it :)

Cargo.toml
```diff
@@ -22,6 +22,8 @@ embedded-hal = "0.2.3"
 display-interface = "0.4"
 display-interface-i2c = "0.4"
 display-interface-spi = "0.4"
+typenum = "1.12.0"
```
jamwaffles, line 25: Small nit: `generic-array` reexports `typenum` so we don't need this additional dependency here.

src/displaysize.rs
```diff
@@ -1,31 +1,101 @@
...
+
+use super::command::Command;
+use display_interface::{DisplayError, WriteOnlyDataCommand};
+use generic_array::ArrayLength;
+use typenum::{U1024, U192, U360, U384, U512};
```
jamwaffles, line 9:
```suggestion
use generic_array::{
    typenum::{U1024, U192, U360, U384, U512},
    ArrayLength,
};
```

bugadani:
> Sorry it's taken a while to get round to this.

No worries. I've done the requested modifications. Thanks for the feedback :)

jamwaffles:
Thanks!

Issue #85, opened 2019-07-01
TerminalMode 90 and 270 degree rotations incorrect


jamwaffles:
As reported in #47.

In `TerminalMode`, the text isn't rotated properly in these two rotations. Instead, it's just drawn backwards for 90 degrees:

![IMG_20190701_212612](https://user-images.githubusercontent.com/800791/60464839-833dfe80-9c47-11e9-9f74-43ab06b0cc7c.jpg)

And like this for 270 degrees:

![IMG_20190701_212641](https://user-images.githubusercontent.com/800791/60464848-8e912a00-9c47-11e9-9210-9ee4683fc071.jpg)

0 and 180 degrees work fine.

bugadani:
Are graphics mode rotations correct? As far as I can tell by looking at the code, only the x and y coordinates are flipped, so I think the displayed image is horizontally flipped?

jamwaffles:
Closed by #129 I believe.

PR #129, opened 2020-06-13
Fix terminal mode

bugadani:

## PR description

~I had a go with #85 and failed miserably. But while doing so, I discovered the driver crashes with 90° rotation. This PR fixes that crash, but does nothing to fix the rotation issue.~

The issue is, properties.get_dimension flips width and height but set_draw_area expects them in the original order. I think for clearing it doesn't matter which direction we go, so I'm using the original physical size instead.

This is a performance hit, because every cursor move sends several display commands, but all the rotation values are rendered correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Changed
 
+- **(breaking)** [#129](https://github.com/jamwaffles/ssd1306/pull/129) `TerminalMode::set_rotation` now resets the cursor position
 - **(breaking)** [#125](https://github.com/jamwaffles/ssd1306/pull/125) Redesigned display size handling.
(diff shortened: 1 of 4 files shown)


jamwaffles:
This fixes the clear issue, but the `terminal_i2c` demo crashes if I add `.with_rotation(DisplayRotation::Rotate90)` or `.with_rotation(DisplayRotation::Rotate270)` to it with this odd result:

image

I haven't had time to investigate but it would be great to fix this new crash in this PR as well, even if the characters remain garbled.

bugadani:
Thanks I forgot to fill my screen to see if there's anything more fishy. Current code writes in the correct direction but newline handling is still incorrect and characters are rotated.

bugadani:
I've fixed the whole thing instead. Code quality could use some love, but hey, priorities :)

Closes #85

therealprof:
> @therealprof you were the original author of `TerminalMode` so I'll let you approve and merge this PR if you're happy with it.

I was afraid you would say that. 😅

bugadani:
reset_pos still needs to be checked, and the docs updated, so take your time :)

bugadani:
Once CI is happy, this is ready for review

jamwaffles:
Awesome. I'll do a release later this evening. Thanks for your patience all!

PR #144, opened 2021-01-08
Exclude various files from the cargo package

### LINKED ISSUE #143: memory.x & build.rs in the library's crate
- ssd1306 = "0.5"
- STM32F411CEUx, but I suppose any MC with RAM size different from 20 KiB is affected
- Display resolution and interface: not applicable

Seems like the provided `memory.x` file overrides the user-provided `memory.x` file when building a binary using the cortex-m-rt crate. Do you think it might be possible to exclude both `build.rs` and `memory.x` files from the crate's library package?

## Test case

Add `ssd1306 = "0.5"` to the target's `Cargo.toml`

```rust
#[cortex_m_rt::entry]
fn main() -> ! {
    const LARGE_ARRAY: [u8; 40 * 1024] = [0; 40 * 1024];
    loop {}
}
```

Expect a hardfault. One might need to use `flip-link` to actually see the hardfault when the seemingly segmentation fault occurs.

Removing `ssd1306` from dependencies resolves the problem.

It seems like the `link.x` script provided by the `cortex-m-rt` crate simply includes any `memory.x` found, and since the `build.rs` you provide adds a viable link search path, the user-defined `memory.x` is ignored. It can be simply proved by removing the user's `memory.x` and observe the compilation/linking succeeds.

mexus:
Remove `memory.x`, `build.rs` and various binary files from the published crate. A side effect is that size of the bundle is decreased down to 35 KB from 2.4 MB.

Closes #143

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,8 @@ repository = "https://github.com/jamwaffles/ssd1306"
 version = "0.5.0"
 edition = "2018"
 
+exclude = [ "build.rs", "build.sh", "memory.x", "doc", "*.jpg", "*.png", "*.bmp" ]
+
 [badges]
 circle-ci = { repository = "jamwaffles/ssd1306", branch = "master" }
 


therealprof (commented):
I'd be careful with some of the image formats since those could potentially directly included but in general that seems like a good idea.

jamwaffles (approved):
Looks good to me. If @therealprof approves, I'll get this merged.

mexus:
> I'd be careful with some of the image formats since those could potentially directly included but in general that seems like a good idea.

The only includes I've managed to identify belong to either examples or build.rs:

```
$ rg include src/
src/mode/terminal.rs
43:/// to include number of characters that can fit on the display.

src/displaysize.rs
13:/// This includes resolution, offset and framebuffer size.
```

therealprof (approved):
Alrighty then.

jamwaffles:
This is released in 0.5.1. Please upgrade and let us know if there are any problems!

mexus:
Yup, it works as expected, thanks a lot!!

PR #158, opened 2021-07-06
Get rid of generic-array

thalesfragoso:

## PR description

`generic-array` really inhibits compiler optimizations, especially when creating one. And since we have big buffers using `GenericArray`, it does make a really big difference. I was having trouble with stack overflows on my 4kB RAM chip, so I went to measure RAM usage, `do_thing` is a function that creates a display driver and prints an simple image to the display:

Current release:
```
0x08000d59      4208    f0_embassy::do_thing::h6248afff2c8c745f
```

With this PR:
```
0x08000d59      1192    f0_embassy::do_thing::h8f7d94a24b293d88
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- **(breaking)** [#158](https://github.com/jamwaffles/ssd1306/pull/158) Migrate away from `generic-array` to a solution using const generics.
+
 ## [0.6.0] - 2021-06-22
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Changed
 
-- **(breaking)** [#158](https://github.com/jamwaffles/ssd1306/pull/158) Migrate away from `generic-array` to a solution using const generics.
+- **(breaking)** [#158](https://github.com/jamwaffles/ssd1306/pull/158) Migrate away from `generic-array` to a solution using const generics. This raises the crate MSRV to 1.51.
 
 ## [0.6.0] - 2021-06-22
 
--- a/src/size.rs
+++ b/src/size.rs
@@ -2,10 +2,18 @@
 
 use super::command::Command;
 use display_interface::{DisplayError, WriteOnlyDataCommand};
-use generic_array::{
-    typenum::{U1024, U192, U360, U384, U512},
-    ArrayLength,
-};
+
+/// Workaround trait, since `Default` is only implemented to arrays up to 32 of size
+pub trait NewZeroed {
+    /// Creates a new value with its memory set to zero
+    fn new_zeroed() -> Self;
+}
+
+impl<const N: usize> NewZeroed for [u8; N] {
+    fn new_zeroed() -> Self {
+        [0u8; N]
+    }
+}
 
 /// Display information.
 ///
@@ -32,7 +40,7 @@ pub trait DisplaySize {
 
     /// Size of framebuffer. Because the display is monocrome, this is
     /// width * height / 8
-    type BufferSize: ArrayLength<u8>;
+    type Buffer: AsMut<[u8]> + NewZeroed;
 
     /// Send resolution and model-dependent configuration to the display
     ///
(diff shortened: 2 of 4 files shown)


therealprof (approved):
Sweet!

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- **(breaking)** [#158](https://github.com/jamwaffles/ssd1306/pull/158) Migrate away from `generic-array` to a solution using const generics.
```
jamwaffles, line 11: I think 1.51 is the right version?

```suggestion
- **(breaking)** [#158](https://github.com/jamwaffles/ssd1306/pull/158) Migrate away from `generic-array` to a solution using const generics. This raises the crate MSRV to 1.51.
```

src/size.rs
```diff
@@ -2,10 +2,18 @@
 
 use super::command::Command;
 use display_interface::{DisplayError, WriteOnlyDataCommand};
-use generic_array::{
-    typenum::{U1024, U192, U360, U384, U512},
-    ArrayLength,
-};
+
+/// Workaround trait, since `Default` is only implemented to arrays up to 32 of size
+pub trait NewZeroed {
```
jamwaffles, line 7: This might be better as `pub(crate)` as AFAIK it's only used internally.

```suggestion
pub(crate) trait NewZeroed {
```

src/size.rs
```diff
@@ -48,7 +56,7 @@ pub struct DisplaySize128x64;
 impl DisplaySize for DisplaySize128x64 {
     const WIDTH: u8 = 128;
     const HEIGHT: u8 = 64;
-    type BufferSize = U1024;
+    type Buffer = [u8; 1024];
```
jamwaffles, line 59: Any reason to not use `Self::WIDTH` and `Self::HEIGHT` here?

```suggestion
    type Buffer = [u8; Self::WIDTH as usize * Self::HEIGHT as usize / 8];
```

I'm looking forward to associated type defaults!

src/size.rs
```diff
@@ -61,7 +69,7 @@ pub struct DisplaySize128x32;
 impl DisplaySize for DisplaySize128x32 {
     const WIDTH: u8 = 128;
     const HEIGHT: u8 = 32;
-    type BufferSize = U512;
+    type Buffer = [u8; 512];
```
jamwaffles, line 72:
```suggestion
    type Buffer = [u8; Self::WIDTH as usize * Self::HEIGHT as usize / 8];
```

src/size.rs
```diff
@@ -74,7 +82,7 @@ pub struct DisplaySize96x16;
 impl DisplaySize for DisplaySize96x16 {
     const WIDTH: u8 = 96;
     const HEIGHT: u8 = 16;
-    type BufferSize = U192;
+    type Buffer = [u8; 192];
```
jamwaffles, line 85:
```suggestion
    type Buffer = [u8; Self::WIDTH as usize * Self::HEIGHT as usize / 8];
```

src/size.rs
```diff
@@ -89,7 +97,7 @@ impl DisplaySize for DisplaySize72x40 {
     const HEIGHT: u8 = 40;
     const OFFSETX: u8 = 28;
     const OFFSETY: u8 = 0;
-    type BufferSize = U360;
+    type Buffer = [u8; 360];
```
jamwaffles, line 100:
```suggestion
    type Buffer = [u8; Self::WIDTH as usize * Self::HEIGHT as usize / 8];
```

src/size.rs
```diff
@@ -105,7 +113,7 @@ impl DisplaySize for DisplaySize64x48 {
     const HEIGHT: u8 = 48;
     const OFFSETX: u8 = 32;
     const OFFSETY: u8 = 0;
-    type BufferSize = U384;
+    type Buffer = [u8; 384];
```
jamwaffles, line 116:
```suggestion
    type Buffer = [u8; Self::WIDTH as usize * Self::HEIGHT as usize / 8];
```

jamwaffles (changes requested):
Thanks for the fix! Just some small nits/comments.

I'll release this as 0.7.0 as the MSRV has increased, along with a minor breaking change to `DisplaySize`.

src/size.rs
jamwaffles: Ah, ignore this I should've tested before commenting... doesn't work.

thalesfragoso: We could sealed it, but not sure if worth it. Not sealing it might be useful if an user wants to write a buffer wrapper for some reason, although not sure if someone would want that...

jamwaffles (approved):
Thanks!

jamwaffles:
[0.7.0 is out](https://crates.io/crates/ssd1306/0.7.0).

PR #161, opened 2021-12-30
Add mirror (vertical flip) option

mryndzionek:

## PR description

Added a function for mirroring/flipping the display. Not that elegant, but seems to be working (tested on one display so far). Let me know if this is acceptable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,6 +314,33 @@ where
         Ok(())
     }
 
+    /// Set mirror/vertical flip.
+    pub fn set_mirror(&mut self, mirror: bool) -> Result<(), DisplayError> {
+        if mirror {
+            match self.rotation {
+                DisplayRotation::Rotate0 => {
+                    Command::SegmentRemap(false).send(&mut self.interface)?;
+                    Command::ReverseComDir(true).send(&mut self.interface)?;
+                }
+                DisplayRotation::Rotate90 => {
+                    Command::SegmentRemap(true).send(&mut self.interface)?;
+                    Command::ReverseComDir(true).send(&mut self.interface)?;
+                }
+                DisplayRotation::Rotate180 => {
+                    Command::SegmentRemap(true).send(&mut self.interface)?;
+                    Command::ReverseComDir(false).send(&mut self.interface)?;
+                }
+                DisplayRotation::Rotate270 => {
+                    Command::SegmentRemap(false).send(&mut self.interface)?;
+                    Command::ReverseComDir(false).send(&mut self.interface)?;
+                }
+            };
+        } else {
+            self.set_rotation(self.rotation)?;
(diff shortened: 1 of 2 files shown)


jamwaffles:
Thanks for the PR! I'll give this a test on my sample displays when I've got a spare minute :)

src/lib.rs
```diff
@@ -314,6 +314,33 @@ where
...
+                    Command::ReverseComDir(true).send(&mut self.interface)?;
+                }
+                DisplayRotation::Rotate180 => {
+                    Command::SegmentRemap(true).send(&mut self.interface)?;
+                    Command::ReverseComDir(false).send(&mut self.interface)?;
+                }
+                DisplayRotation::Rotate270 => {
+                    Command::SegmentRemap(false).send(&mut self.interface)?;
+                    Command::ReverseComDir(false).send(&mut self.interface)?;
+                }
```
jamwaffles, line 336: In my testing, I found the 270º and 90º configs should be swapped:

```suggestion
            match self.rotation {
                DisplayRotation::Rotate0 => {
                    Command::SegmentRemap(false).send(&mut self.interface)?;
                    Command::ReverseComDir(true).send(&mut self.interface)?;
                }
                DisplayRotation::Rotate90 => {
                    Command::SegmentRemap(false).send(&mut self.interface)?;
                    Command::ReverseComDir(false).send(&mut self.interface)?;
                }
                DisplayRotation::Rotate180 => {
                    Command::SegmentRemap(true).send(&mut self.interface)?;
                    Command::ReverseComDir(false).send(&mut self.interface)?;
                }
                DisplayRotation::Rotate270 => {
                    Command::SegmentRemap(true).send(&mut self.interface)?;
                    Command::ReverseComDir(true).send(&mut self.interface)?;
                }
            };
```

Did you see different results on your display? Can you apply this suggestion and retest on your end? Hopefully we don't have different behaviour!

src/lib.rs
```diff
@@ -314,6 +314,33 @@ where
         Ok(())
     }
 
+    /// Set mirror/vertical flip.
```
jamwaffles, line 317: I think just mentioning mirroring behaviour is fine:

```suggestion
    /// Set mirror enabled/disabled.
```

jamwaffles (changes requested):
Looks good so far, just some changes from my testing.

Also needs a changelog entry, something like:

```markdown
### Added

- [#161](https://github.com/jamwaffles/ssd1306/pull/161) Added a `set_mirror` method to enable or disable display mirroring.
```

added after the `Unreleased` heading.

jamwaffles (approved):
Apologies, I didn't see your updates until now! Please throw me a comment for any future PRs so I don't forget :). Thanks for the fix!

PR #176, opened 2023-05-02
Simplify terminal mode error handling

bugadani:
While coherence rules don't allow us to implement `From<Result<...>>` we can do `From<OneError> for OtherError` and let the `?` operator handle the conversion.

I don't believe this needs a changelog entry but let me know if it does.


jamwaffles (approved):
Much cleaner, thank you!

This doesn't need a changelog entry - it's just internal work :)

PR #179, opened 2023-05-02
Remove a generic param, signify no comm error during reset

bugadani:

## PR description

Reset can't return a comm error, so let's make it Infallible. Also, there's no need for a third generic type.


jamwaffles (approved):
Noice

PR #180, opened 2023-05-02
Move interface-independent code out into new impl blocks

bugadani:

## PR description

Some code doesn't rely on the display interface. This PR moves these to impl blocks that don't restrict DI to any particular trait. While this isn't immediately useful, this refactor will help in avoiding code duplication.

I've added `FastNoop` instead of short circuiting the current `Noop` command just in case someone relies on the current command's details. [Backwards compatibility](https://xkcd.com/1172/) FTW. `FastNoop` is currently used to fill out Command arrays where a particular option requires less elements than the array's capacity.

Warning: some of my decisions may be controversial.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -87,16 +87,16 @@ pub enum Command {
     ChargePump(bool),
     /// Select external or internal I REF. Only for 72 x 40 display with SSD1306B driver
     InternalIref(bool, bool),
+    /// Noop that isn't even sent through the interface.
+    ///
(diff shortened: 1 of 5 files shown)


src/command.rs
```diff
@@ -8,7 +8,6 @@ use display_interface::{DataFormat::U8, DisplayError, WriteOnlyDataCommand};
 
 /// Commands
 #[derive(Debug, Copy, Clone)]
-#[allow(dead_code)]
```
bugadani, line 11: Drive-by delete because why not

PR #181, opened 2023-05-14
Bump e-g core

bugadani:
Hi! Thank you for helping out with SSD1306 development! Please:

- [x] Check that you've added documentation to any new methods
- [x] Rebase from `master` if you're not already up to date
- [x] Add or modify an example if there are changes to the public API
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, **Changed**, etc)
- [x] Run `rustfmt` on the project with `cargo fmt --all` - CI will not pass without this step
- [x] Check that your branch is up to date with master and that CI is passing once the PR is opened

## PR description

Update embedded-graphics and modernize examples.

PR #184, opened 2023-05-16
Modernize examples

bugadani:
This PR updates examples to the currently latest HAL/RTIC crates, making them relevant again. Hopefully, this provides a solid base for later embedded-hal 1.0 updates, too.

As I don't have an STM32F10x, I haven't been able to test them, but I'm confident in everything except the RTIC examples.


bugadani:
Merged into #181

PR #185, opened 2023-05-18
Some general cleanup

### LINKED ISSUE #174: Using `DrawTarget::clear` is difficult in buffered graphics mode
- Version of `ssd1306` in use (if applicable): 0.71

There are two different `clear` methods for `Ssd1306<DI, SIZE, BufferedGraphicsMode<SIZE>>`: One in an `impl` (https://docs.rs/ssd1306/latest/src/ssd1306/mode/buffered_graphics.rs.html#66-191) and the other in the `DrawTarget` trait. The non trait `impl` seems to take precedence and the compiler doesn't even output an error message that multiple `clear` methods exist, like it would when both `clear` methods would be declared in traits.

I'm not sure what the best solution to this is. From an e-g standpoint I would suggest to remove the non `DrawTarget` `clear` method, because a user might expect clearing to work like in the examples (https://github.com/embedded-graphics/examples/blob/main/eg-0.7/examples/demo-analog-clock.rs#L195) or other drivers (https://github.com/almindor/mipidsi#example). But that would create some inconsistency between the different modes.

## Test case (if applicable)

```rust
display.clear(BinaryColor::On).unwrap(); 
```

```
error[E0061]: this function takes 0 arguments but 1 argument was supplied
  --> examples/hello-world.rs:88:15
   |
88 |     ssd1306_0.clear(BinaryColor::On).unwrap();
   |               ^^^^^ --------------- argument of type `BinaryColor` unexpected

```

This works as expected:
```rust
DrawTarget::clear(&mut ssd1306_0, BinaryColor::On).unwrap();
```

bugadani:

## PR description

This PR is a set of smaller changes extracted from the other pending pull requests, that are otherwise independent of the bigger changes proposed.

A few notes:

 - Fixes #174 by only providing either the DrawTarget's `clear`, or the inherent one, but not both. I've chosen not to add a clear color to the inherent method to keep it the same as before.
 - Clearing the display on an ESP32 before this patch took 1.4ms, measured with a tracing pin and an oscilloscope. After, this time got down to around 100us but I haven't precisely measured that, I was happy with seeing just the relative difference.
 - I'm deduplicating the reset fn as it's the same for all display interfaces.
 - There's `allow(dead_code)` on a type that doesn't have dead code warnings.


src/mode/buffered_graphics.rs
```diff
@@ -69,9 +69,14 @@ where
...
+    #[cfg(not(feature = "graphics"))]
     pub fn clear(&mut self) {
+        self.clear_impl(false);
+    }
+
+    fn clear_impl(&mut self, value: bool) {
         for b in self.mode.buffer.as_mut() {
-            *b = 0;
+            *b = value as u8;
         }
```
rfuest, line 80: Isn't this the same as `self.mode.buffer.fill(value as u8)`? Maybe that's faster if the compiler didn't already optimize the original code into the same thing.

bugadani: You're right, this is some sort of fill. On the other hand, this doesn't matter in term of performance as much as not using `fill_solid` is.

bugadani:
> My opinion on this matter might not be important, but I really don't like the idea of removing a method when a feature is added.

I'm interested in other ways of resolving the conflict while preserving the ability to clear the display without embedded-graphics, if you have any.

rfuest:
My solution would be to just require e-g for the buffered graphics mode. I think that most users will use e-g anyway and it isn't a huge dependency if you don't want to use e-g and just set pixels. But I'm not sure everyone would like this approach.

rfuest:
If the buffered mode is changed to use e-g internally (see https://github.com/jamwaffles/ssd1306/pull/169) e-g will become a mandatory dependency anyway.

bugadani:
My concern is that requiring e-g for the buffered graphics mode will require other potential graphics libraries to implement a compatibility display mode for this driver.

While e-g's framebuffer is a good general idea, this display is pretty much designed to work with a framebuffer, or at least to draw rectangles of pixels to it. I expect a built-in framebuffer to remain useful without embedded-graphics, especially considering the current status of the library.

The underlying issue in my opinion is the poor naming of `DrawTarget::clear`, which should probably been called `fill`. `clear` to me suggest a return to a default state, exactly how the inherent method does here. That the default state isn't configurable, is a side-effect of the display being a monochrome OLED.

While I agree that removing methods by enabling a feature is inconvenient, and has a sizeable surprise factor, the only other viable solution I see is to rename it. But I would prefer other solutions as I feel like `clear` is... well, pun not intended, pretty clear.

rfuest:
> The underlying issue in my opinion is the poor naming of `DrawTarget::clear`, which should probably been called `fill`. `clear` to me suggest a return to a default state, exactly how the inherent method does here. That the default state isn't configurable, is a side-effect of the display being a monochrome OLED.

The name `clear` was chosen to match similar operations in other APIs, like OpenGL and Vulkan.

bugadani:
Considering that `clear` just cleared the framebuffer and not the actual display, I've changed my opinion and renamed it to `clear_buffer`. Hopefully this is both a good compromise and a change that will make the function's behaviour clearer.

jamwaffles (approved):
Thanks 🙏

PR #189, opened 2023-07-07
[WIP] embedded-hal/async update

bugadani:
Will replace #178

note to self: buffered mode is missing async `init()` (`DisplayConfig` is sync only)


eldruin:
Superseeded by #213  and #214.
Thank you for your work!

PR #195, opened 2023-09-20
Optimise BasicMode::clear RAM usage and optimize size of Ssd1306::char_to_bitmap

alexey-medvedchikov:

- [X] Rebase from `master` if you're not already up to date
- [X] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, **Changed**, etc)
- [X] Run `rustfmt` on the project with `cargo fmt --all` - CI will not pass without this step
- [X] Check that your branch is up to date with master and that CI is passing once the PR is opened

## PR description

This PR fixes issues with small AVRs:
* BasicMode doesn't work on MCUs with less than ~1.5Kb of RAM because there's a stack array of 1024 bytes allocated to zero-out pages. Instead, we can zero-out in small pieces.
* Refactor glyph search to use lookup table instead of one big `match`. Saves around 3.5Kb of compiled code on AVR ATMega-s. Didn't test on other architectures, but pretty sure it will save some size too.

### AVR size difference

Before:
```
text    data     bss     dec     hex filename
   0    8446       0    8446    20fe object.hex
```

After:
```
text    data     bss     dec     hex filename
   0    4988       0    4988    137c object.hex
```

### STM32 (thumbv7em) size difference

Before:
```
verified 7212 bytes
```

After:
```
verified 6392 bytes
```


src/mode/mod.rs
```diff
@@ -31,11 +31,14 @@ where
...
-        self.set_draw_area((0, 0), self.dimensions())?;
+        let dim = self.dimensions();
+        self.set_draw_area((0, 0), dim)?;
 
-        // TODO: If const generics allows this, replace `1024` with computed W x H value for current
-        // `SIZE`.
-        self.draw(&[0u8; 1024])
+        for _ in 0..dim.0 * dim.1 {
+            self.draw(&[0; 8])?;
+        }
```
jamwaffles, line 39: I've got a couple of questions here. I hope you don't mind a short list:

1. Because you're sending 8 bytes in a row, shouldn't this be `0..(dim.0 * dim.1) / 8`?
2. Could we increase the slice size to 64 or 128 bytes? This will still save a decent amount of memory but will reduce the overhead of doing the CS pin setup, etc.
3. To add to (1.), I think you might need a further `/ 8` because the display's dimensions are in pixels, but a page is 8 bytes tall.

jamwaffles (changes requested):
Hey, thanks for the improvements! The memory size reductions are impressive. I left a few questions on your `clear()` changes.

I also want to test this PR on some hardware before merging but I'm travelling for this week. I hope you don't mind the delay.

alexey-medvedchikov:
@jamwaffles No problemo!
Although I saved some space, I still can't squeeze my program into atmega168 - bme680 driver is a resource hog. So I'll try to switch to a STM32 MCU, I can test this change there too later this week. I also have ESP8266 at hand, but I've never used it with Rust, so not sure if I can quickly check it.

src/mode/mod.rs
alexey-medvedchikov: Addressed the comments you provided. Also, I realised that the addressing mode must be set and restored, so did that too.

alexey-medvedchikov:
LOL, I've just noticed that all the first and last bytes of each glyph in the font are zeros, so I just made the data array smaller by another 500 bytes. It is a bit clunky, but the font is hardcoded (and quite nice), so I don't think it is a problem. -3.5Kb total.

alexey-medvedchikov:
Tested on NUCLEO-F446ZE and ATMega168 with 128x64 and 128x32, all work fine.

alexey-medvedchikov:
Got rid of extra-sized arrays in `Command::send`, cuts 80 bytes on Cortex-M4 and 260 bytes on ATMega168.

src/mode/mod.rs
```diff
@@ -31,11 +31,32 @@ where
...
+        let old_addr_mode = self.addr_mode;
+        if old_addr_mode != AddrMode::Horizontal {
+            self.set_addr_mode(AddrMode::Horizontal)?;
+        }
 
-        // TODO: If const generics allows this, replace `1024` with computed W x H value for current
-        // `SIZE`.
-        self.draw(&[0u8; 1024])
+        let dim = self.dimensions();
+        self.set_draw_area((0, 0), dim).unwrap();
```
jamwaffles, line 40: This can be an early return on error - no need to panic. Alas no reduction in binary size :(

```suggestion
        self.set_draw_area((0, 0), dim)?;
```

src/mode/terminal.rs
```diff
@@ -302,104 +302,206 @@ where
...
+            [0x02, 0x01, 0x02, 0x01, 0x00, 0x00],
+            // blank
+            [0x00, 0x00, 0x00, 0x00, 0x00, 0x00],
+        ];
+
+        let g = if ('!'..='~').contains(&input) {
+            CHARS[(input as u8 - b'!') as usize]
+        } else {
+            CHARS[CHARS.len() - 1]
+        };
```
jamwaffles, line 502: This could be written a bit more elegantly like this:

```suggestion
        let g = CHARS
            .get((input as u8 - b'!') as usize)
            .unwrap_or(&CHARS[CHARS.len() - 1]);
```

It also saves a few bytes in my testing but YMMV! Can you confirm a size reduction in your setup?

src/mode/terminal.rs
```diff
@@ -302,104 +302,206 @@ where
...
-            'x' => [0x00, 0x22, 0x14, 0x08, 0x14, 0x22, 0x00, 0x00],
-            'y' => [0x00, 0x1c, 0xa0, 0xa0, 0xa0, 0x7c, 0x00, 0x00],
-            'z' => [0x00, 0x22, 0x32, 0x2a, 0x26, 0x22, 0x00, 0x00],
-            '{' => [0x00, 0x0c, 0x3f, 0x21, 0x00, 0x00, 0x00, 0x00],
-            '|' => [0x00, 0x3f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00],
-            '}' => [0x00, 0x21, 0x3f, 0x0c, 0x00, 0x00, 0x00, 0x00],
-            '~' => [0x00, 0x02, 0x01, 0x02, 0x01, 0x00, 0x00, 0x00],
-            _ => [0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00],
-        }
+        const CHARS: [[u8; 6]; 95] = [
```
jamwaffles, line 305: Nice optimisation! In hindsight it was silly to store a pile of extra zeros but thanks for the fix!

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- [#195](https://github.com/jamwaffles/ssd1306/pull/195) Changed `BasicMode::clear` to clear in small batches instead of
+  one big write. This drops RAM requirement by ~900b and fixes issues on MCUs with less than 1Kb of RAM.
+- [#195](https://github.com/jamwaffles/ssd1306/pull/195) Changed `TerminalMode` to use lookup by ascii code instead of
```
jamwaffles, line 13: Nit :)

```suggestion
- [#195](https://github.com/jamwaffles/ssd1306/pull/195) Changed `TerminalMode` to use lookup by ASCII code instead of
```

jamwaffles (changes requested):
Thanks very much for testing the different display sizes - that's saved me some time :). I've left a couple of inline suggestions which make the code a bit cleaner. Let me know what you think.

alexey-medvedchikov:
Thanks for the review. I accepted all your suggestions, a newbie in Rust, so make simple mistakes. Also, the merge conflict should be resolved now.

src/mode/terminal.rs
alexey-medvedchikov: Comitted this, but reverted because if input is smaller than '!' (for example space or newline) it leads to a panic: https://app.circleci.com/pipelines/github/jamwaffles/ssd1306/620/workflows/40c25c3e-6806-4af1-a7e7-679410eb9579/jobs/5370

jamwaffles: Right, sorry about that - I wasn't able to run the tests when I made this suggestion. How about this:

```suggestion
        let g = (input as usize)
            .checked_sub(b'!'.into())
            .and_then(|idx| CHARS.get(idx))
            .unwrap_or(&CHARS[CHARS.len() - 1]);
```

The `checked_sub` returns `None` on underflow, so won't panic any more. I ran a few tests and this seems to work pretty nicely. It also still helps the binary size compared to the original code which is good.

jamwaffles (changes requested):
Hey, sorry for the review delay. I've got one more nit that hopefully doesn't fail the tests any more, then this is good for merge. Thanks for your patience!

src/mode/terminal.rs
alexey-medvedchikov: Committed. Thanks, TIL!

I've checked the x86 assembly output for my version vs your version and your version is definitely more compact. Is it because this way we express intent more clearly and rust compiler is capable to trigger some extra optimisation techniques?

alexey-medvedchikov:
@jamwaffles checked and committed your changes. Thanks for spending so much time on this one!

src/mode/terminal.rs
jamwaffles: > Is it because this way we express intent more clearly and rust compiler is capable to trigger some extra optimisation techniques?

Yeah, something like that :D. I don't have a very deep understanding of how all this works at the low level but a rule of thumb I follow is if I can chain some methods instead of doing it imperitively (`if` statements, etc), Rust will _likely_ be able to do more optimisations with them and create faster code as it can make more assumptions, or can discard invariants it knows can't exist any more, etc.

jamwaffles (approved):
Looks good! Thanks again for your patience :)

jamwaffles:
This is released in v0.8.3. `cargo update` should pull in the changes.

bugadani:
FYI I'm firmly opposed to https://github.com/jamwaffles/ssd1306/pull/195/commits/e309b41027e4219a17ba28200034eddcff25da77 as it means I'll have to duplicate the whole command table for the async interface.

PR #197, opened 2023-09-28
Fix rotated terminal panic/behaviour

### LINKED ISSUE #196: Using terminal mode with 90/270 degree rotation leads to incorrect output and panic
- Version of `ssd1306` in use (if applicable): 0.8.1
- MCU/other hardware in use: Seeeduino XIAO RP2040 with expansion board
- Display resolution and interface: I2C, 128x64

When `TerminalMode` and `Rotation90` or `Rotation270` is used, writing a new line of text results in the first character being written to the first line of the display, with the remainder of the text for the same line displayed on a later, but expected, line. After writing the eighth line, attempts to write the ninth line trigger a panic, possibly because the first character of the ninth line would (incorrectly) need to be written off screen.

## Test case (if applicable)

A sample reproducer using `rp2040-hal` is provided in [plm/ssd1306-rotation-panic](https://github.com/plm/ssd1306-rotation-panic), with the relevant lines below.

```rust
let i2c = ... // omitted for brevity
let mut display = I2CDisplayInterface::new(i2c),
    DisplaySize128x64,
    DisplayRotation::Rotate270,
)
.into_terminal_mode();

if display.init().is_ok() && display.clear().is_ok() {
   let _ = writeln!(display, "INIT OK");
   loop {
       for count in 0..u8::MAX {
           let _ = writeln!(display, "{}TEST", count); // panics on count == 7
       }
   }
}
```

Expected display output, which should repeat indefinitely:

```
INIT OK
0TEST
1TEST
2TEST
3TEST
4TEST
5TEST
6TEST
7TEST
```

Actual display output, which panics when writing the eighth test line:

```
I0123456
 TEST
 TEST
 TEST
 TEST
 TEST
 TEST
 TEST
```

jamwaffles:
Closes #196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ SSD1306 monochrome OLED display.
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- [#197](https://github.com/jamwaffles/ssd1306/pull/197) Fixed terminal mode panic and wrapping
+  behaviour for rotated displays.
+
 ## [0.8.1] - 2023-08-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
-[`ssd1306`](https://crates.io/crates/ssd1306) is a no-std driver written in Rust for the popular SSD1306 monochrome OLED display.
+[`ssd1306`](https://crates.io/crates/ssd1306) is a no-std driver written in Rust for the popular
+SSD1306 monochrome OLED display.
 
 <!-- next-header -->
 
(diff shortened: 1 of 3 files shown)


jamwaffles:
Tested on 128x64 display with `Rotation270` and it works nicely for me.

plm:
Thank you for the quick fix @jamwaffles! I can confirm the fix works in the [`fixed` branch of my reproducer](https://github.com/plm/ssd1306-rotation-panic/tree/fixed).

jamwaffles:
Great! Thanks for testing! This is released in [0.8.2](https://crates.io/crates/ssd1306/0.8.2).

PR #207, opened 2024-01-15
Update to embedded-hal-1.0.0

irii:
Hi! Thank you for helping out with SSD1306 development! Please:

- [x] Check that you've added documentation to any new methods
- [x] Rebase from `master` if you're not already up to date
- [x] Add or modify an example if there are changes to the public API
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, **Changed**, etc)
- [x] Run `rustfmt` on the project with `cargo fmt --all` - CI will not pass without this step
- [ ] Check that your branch is up to date with master and that CI is passing once the PR is opened

## PR description

Updated to embedded-hal 1.0.0
Updated rust-version to 1.75 for display-interface package

- > Works on an CM4 module with I2C

PR #212, opened 2024-06-16
Adjust for r e c

sjoerdsimons:

## PR description

Various updates now it's hosted on r-e-c:
* Switch from circleci to github actions
* Adjust links to point at r-e-c repository
* Fix issues point out by newer rustdoc and rustc


sjoerdsimons:
Successful github actions can be found in my repo :)

memory.x
```diff
@@ -1,6 +1,6 @@
 /* Linker script for the STM32F103C8T6 */
 MEMORY
 {
-  FLASH : ORIGIN = 0x08000000, LENGTH = 128K
+  FLASH : ORIGIN = 0x08000000, LENGTH = 64K
```
eldruin, line 4: Is there a reason to change this?

eldruin (commented):
Nice, thank you!
Could you also update the build badge at the top of the readme?

memory.x
sjoerdsimons: Fwiw independantly ran into a note that there are some variants of the stm32f103 on some blue pills that do have 128K ; so i wonder if @jamwaffles has one of those ;) .. the STM32F103C8T6 officially only has 64K though

jamwaffles: That's exactly it! I don't think I own a single non-bootleg F103 :grin: 

FWIW this setup is way out of date. I believe these days one doesn't need to provide their own `memory.x`, so this file could probably be deleted.

sjoerdsimons: not with stm32f1xx-hal afaik; i'm prepping a follow-up PR that moves the examples to embassy for the pac (as that implements hal 1.x) together with the patches from #207 so that that can be merged hopefully.. That will also drop memory.x

eldruin (approved):
Looks good to me, thank you!

sjoerdsimons:
@eldruin  thanks for merging! Looks like someone needs to disable the circleci app such that github declares CI as succeeding (currently it fails because there is no circleci config...)

PR #214, opened 2024-08-10
Async API

sjoerdsimons:

## PR description

Add async support to crate; API stays the same just async

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,9 @@ jobs:
         with:
             toolchain: 1.75
       - run: cargo build --lib --target x86_64-unknown-linux-gnu
+      - run: cargo build --lib --target x86_64-unknown-linux-gnu --features async
       - run: cargo doc --target x86_64-unknown-linux-gnu
+      - run: cargo doc --target x86_64-unknown-linux-gnu --features async
 
   build:
     strategy:
@@ -68,6 +70,8 @@ jobs:
           components: rustfmt
       - run: rustup target add ${{matrix.target}}
       - run: cargo build --target ${{matrix.target}} --all-features --release
+      - if: ${{ matrix.examples }}
+        run: cargo build --target ${{matrix.target}} --examples --release
       - if: ${{ matrix.examples }}
         run: cargo build --target ${{matrix.target}} --all-features --examples --release
       - run: cargo doc --all-features --target ${{matrix.target }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ SSD1306 monochrome OLED display.
 
 - Updated dependencies for `embedded-hal` 1.0.0.
 - Switch examples to embassy STM32 PAC which implements `embedded-hal` 1.0.0 traits.
+- Add an asynchronous interface, enabled via the `async` feature.
 - **(breaking)** Increased MSRV to 1.75.0
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,9 @@ display-interface-spi = "0.5.0"
 embedded-graphics-core = { version = "0.4.0", optional = true }
 
 [dev-dependencies]
+embedded-graphics = "0.8.0"
+
+[target.'cfg(target_arch="arm")'.dev-dependencies]
 cortex-m = { version = "0.7.2", features = ["critical-section-single-core"] }
 cortex-m-rt = "0.7.3"
 cortex-m-rtic = "1.1.4"
@@ -32,7 +35,6 @@ defmt-rtt = "0.4.0"
 panic-probe = { version = "0.3.1", features = ["print-defmt"] }
 # Used to load BMP images in various examples
 tinybmp = "0.5.0"
-embedded-graphics = "0.8.0"
 # Used by the noise_i2c examples
 rand = { version = "0.8.4", default-features = false, features = [ "small_rng" ] }
 embassy-stm32 = { version = "0.1.0", git = "https://github.com/embassy-rs/embassy", features = [ "stm32f103c8", "memory-x", "defmt", "exti", "time-driver-tim3" , "unstable-pac"]  }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ rust-version = "1.75.0"
 
 [package.metadata.docs.rs]
 targets = [ "thumbv7m-none-eabi", "thumbv7em-none-eabihf" ]
+all-features = true
 
 [dependencies]
 embedded-hal = "1.0.0"
@@ -22,6 +23,8 @@ display-interface = "0.5.0"
 display-interface-i2c = "0.5.0"
 display-interface-spi = "0.5.0"
 embedded-graphics-core = { version = "0.4.0", optional = true }
+embedded-hal-async = { version = "1.0.0",  optional = true }
+maybe-async-cfg = "0.2.4"
 
 [dev-dependencies]
 embedded-graphics = "0.8.0"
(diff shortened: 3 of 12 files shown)


sjoerdsimons:
As usual the circleci pipeline files (would be nice if it's removed at some point :p; @eldruin  hope you don't mind me giving a ping as i don't know if you already notifications in new PR's :)

.github/workflows/ci.yml
```diff
@@ -29,15 +29,17 @@ jobs:
       - run: cargo test --doc --target x86_64-unknown-linux-gnu
 
   test-msrv:
-    name: build with MSRV
+    name: test with MSRV
```
eldruin, line 32: I would prefer to only build with the MSRV and test with stable like it is currently done because we have a significant number of `dev-dependencies` and so it is quite probable that these will break the MSRV test although the failure is not related to this library.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ SSD1306 monochrome OLED display.
 
 - Updated dependencies for `embedded-hal` 1.0.0.
 - Switch examples to embassy STM32 PAC which implements `embedded-hal` 1.0.0 traits.
+- Add an async interface, enabled via the async feature.
```
eldruin, line 12:
```suggestion
- Add an asynchronous interface, enabled via the `async` feature.
```

src/lib.rs
```diff
@@ -406,24 +502,34 @@ where
...
+        for c in buffer
             .chunks(disp_width)
             .skip(starting_page)
             .take(num_pages)
             .map(|s| &s[page_lower..page_upper])
-            .try_for_each(|c| interface.send_data(U8(c)))
-    }
-
-    /// Release the contained interface.
-    pub fn release(self) -> DI {
```
eldruin, line 418: why would you remove his method?

src/mode/terminal.rs
```diff
@@ -141,31 +218,44 @@ where
...
     /// Initialise the display in page mode (i.e. a byte walks down a column of 8 pixels) with
     /// column 0 on the left and column _(SIZE::Width::U8 - 1)_ on the right, but no automatic line
     /// wrapping.
-    fn init(&mut self) -> Result<(), TerminalModeError> {
-        self.init_with_addr_mode(AddrMode::Page)?;
-        self.reset_pos()?;
+    async fn init(&mut self) -> Result<(), TerminalModeError> {
+        self.init_with_addr_mode(AddrMode::Page).await?;
+        self.reset_pos().await?;
         Ok(())
```
eldruin, line 233:
```suggestion
        self.reset_pos().await
```

src/mode/terminal.rs
```diff
@@ -178,29 +268,30 @@ where
...
+            self.draw(&[0; 8]).await?;
         }
 
         // But for normal operation we manage the line wrapping
-        self.set_addr_mode(AddrMode::Page)?;
-        self.reset_pos()?;
+        self.set_addr_mode(AddrMode::Page).await?;
+        self.reset_pos().await?;
 
         Ok(())
```
eldruin, line 283:
```suggestion
        self.reset_pos().await
```

src/mode/terminal.rs
```diff
@@ -276,19 +367,19 @@ where
         self.mode.cursor = Some(Cursor::new(w, h));
 
         // Reset cursor position
-        self.set_position(0, 0)?;
+        self.set_position(0, 0).await?;
 
         Ok(())
```
eldruin, line 372:
```suggestion
        self.set_position(0, 0).await
```

src/mode/terminal.rs
```diff
@@ -276,19 +367,19 @@ where
...
-    fn advance_cursor(&mut self) -> Result<(), TerminalModeError> {
+    async fn advance_cursor(&mut self) -> Result<(), TerminalModeError> {
         let cursor = self.ensure_cursor()?;
 
         cursor.advance();
         let (c, r) = cursor.get_position();
-        self.set_position(c, r)?;
+        self.set_position(c, r).await?;
 
         Ok(())
```
eldruin, line 384:
```suggestion
        self.set_position(c, r).await
```

eldruin (commented):
Thank you! Seems fine to me aside from the comments below but I am no async expert. Have you tested this in hardware?

sjoerdsimons:
> Thank you! Seems fine to me aside from the comments below but I am no async expert. Have you tested this in hardware?

Ofcourse; e.g. the async-i2c-spi example running on a breadboard

https://github.com/user-attachments/assets/b95d3b61-6250-4baf-9e46-a207ab5606bd

.github/workflows/ci.yml
sjoerdsimons: Happy to do so; but mind that with these patches the only dev-dependency on x86 left is embedded-graphics itself; Which is a lot more tractable then all the bigs required for arm (which is primarily for the examples)

src/lib.rs
sjoerdsimons: because it got lost while trying some different patterns :)

.github/workflows/ci.yml
sjoerdsimons: @eldruin sorry the previous message was meant to ask if you'd reconsider with that background :) Fwiw the main thing we do test with MSRV in this case is the doctest which imho is useful (and should really work with the MSRV)

If you'd still prefer to just build with MSRV then i'll do that change

eldruin: I understand the utility from a software engineering perspective.
This is a matter of differing expectations w.r.t. what MSRV means. As a user, the MSRV means to me: can I use it as a dependency?
If the doctests do not compile anymore due to changes in MSRV in, say, embassy, does it mean that this crate must change its MSRV? I would say no. The ecosystem around it changed but this crate or its actual dependencies did not.
It should be noted that there are a [whole bunch](https://github.com/rust-embedded-community/ssd1306/actions/runs/10392964317/job/28779387012?pr=214#step:4:1) of dependencies that need to be compiled for this.
What do you think?

sjoerdsimons: Right my point was that with the changes i've done for x86 the only dev-dependency left is embedded-graphics itself (which any users is expected to also use).. So the whole bunch of crates is essentially the same betweenthe test and a build a user will do :)

eldruin: I see. Sorry, I missed that change.

eldruin (commented):
Is this ready to go now?

.github/workflows/ci.yml
sjoerdsimons: Hah; Well i rolled it back now as i didn't see the point in blocking on this for longer :) Happy to send a quick follow-up pr to re-introduce testing in ci if you changed your mind ofcourse

sjoerdsimons:
> Is this ready to go now?

That's for you to judge! but from my point of view yes

Cargo.toml
```diff
@@ -37,13 +40,25 @@ panic-probe = { version = "0.3.1", features = ["print-defmt"] }
 tinybmp = "0.5.0"
 # Used by the noise_i2c examples
 rand = { version = "0.8.4", default-features = false, features = [ "small_rng" ] }
+embassy-embedded-hal = { version = "0.2.0", git = "https://github.com/embassy-rs/embassy" }
```
plaes, line 43: Why git dependency?

sjoerdsimons: Because 0.2 wasn't even on crates.io when most of this was done ;)

I could get rid of the git dependency of embedded-hal (by disabling the default features); embassy-executor still has to come from git as it has to agree with embassy-stm32 for the time driver

plaes: Ah, indeed. I have only tried with embassy-nrf and esp-hal.

eldruin (approved):
Looks good to me, thank you!

plaes:
> Looks good to me, thank you!

Could you bump version as well?

eldruin:
I have just released this in [version 0.9.0](https://github.com/rust-embedded-community/ssd1306/releases/tag/v0.9.0)

PR #232, opened 2025-06-24
perf: implement fill_contiguous

arlosi:
A significant amount of CPU time was being spent filling rectangles in my application. I traced most of the problem to `set_pixel` being somewhat slow for large rectangular areas.

By implementing `fill_contiguous` much of the bounds checking and control flow changes for rotation are lifted outside the hot path.

Tested on STM32G0 with 0.96 128x64 OLED over I2C.


eldruin (approved):
Sounds good to me, thank you!

PR #236, opened 2026-07-04
Reduce unnecessary chunking

bugadani:

## PR description

This PR sends contiguous buffers in one operation, instead of chunking it up in 8-pixel tall stripes. If you update the whole display, the underlying display driver can transfer the buffer much more efficiently - in my case, with a 128x64 display, the driver issues a single send_data call instead of 8.

Before (ignore the 8MHz, my logic analyser is too slow to sample the 40MHz clock so it's aliasing):

image

After

image


MabezDev (approved):
LGTM

